PR #4, opened 2022-08-19
add various nodes

MichalMarsalek:
Polygolf builtins:
* sort -> sorted
* contains_key
* contains_values
* indexof

Polygolf nodes:
* Argv
* VarDeclaration
* ArrayConstructor
* MapSet

Specific nodes:
* Import
* MutatingBinaryOp
* ConditionalOp
* ArrayGet
* MapGet
* ArraySet
* ForRange
* ForEach
* ForEachPair
* ForCLike
* ManyToManyAssignment
* OneToManyAssignment

Others:
* Variable type enum
* Add imports & vars fields to `Program`

### DIFF
--- a/src/IR/IR.ts
+++ b/src/IR/IR.ts
@@ -88,14 +88,6 @@ export type VariableType =
 }
 
 /**
- * Import.
- */
- export interface Import {
-  type: "Import";
-  name: string;
-}
-
-/**
  * A general function application, such as (+ a b) or (print x). Raw OK
  *
  * Every language frontend should convert *all* function applications to
@@ -189,6 +181,14 @@ export interface MapSet {
 }
 
 /// === Interfaces below here are language-specific ===
+
+/**
+ * Import.
+ */
+ export interface Import {
+  type: "Import";
+  name: string;
+}
 
 export interface FunctionCall {
   type: "FunctionCall";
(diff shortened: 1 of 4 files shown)


src/IR/IR.ts
```diff
@@ -18,7 +18,16 @@ export type Expr =
...
-  block: Block;
+  imports: Import[];
+  varDeclarations: VarDeclaration[];
+  block: Block;  
+};
+
+/**
+ * Program input (array of strings).
+ */
+ export type Argv = {
```
jared-hughes, line 29: Argv doesn't seem to be referenced anywhere. Should it be directly on Program?

I'm guessing that it would be used to enable certain optimizations such as assuming argc=1 in C programs.

src/IR/IR.ts
```diff
@@ -64,6 +73,20 @@ export interface Assignment {
   expr: Expr;
 }
 
+export type VariableType =
```
jared-hughes, line 76: VariableType seems to be missing some types (list<?>, table<?,?>). Those are more complicated to represent; maybe we just leave off the types for now; see next comment.

jared-hughes (changes requested):

src/IR/IR.ts
MichalMarsalek: > Argv doesn't seem to be referenced anywhere. Should it be directly on Program?
> 
> I'm guessing that it would be used to enable certain optimizations such as assuming argc=1 in C programs.

It represents a nullary builtin - the program input. It will translate to `sys.argv` for Python for example.

jared-hughes: Oh! That makes sense! It still needs to be added to the `Expr` union though

jared-hughes (approved):

PR #10, opened 2022-08-22
Common plugins

MichalMarsalek:
Fixed `npm run cli`.
Added plugins for converting `ForRange` to all other loop nodes & created tests for them.
Added `addMutatingBinaryOp` plugin.
Added `root` pointing to the Program node's path for each path.
Added `Path.getUsedIdentifiers()`.
Added `Path.getNewIdentifier()`.
Implemented more nodes in the debug emitter.


src/IR/IR.ts
```diff
@@ -271,7 +272,7 @@ export interface BinaryOp {
  */
 export interface MutatingBinaryOp {
   type: "MutatingBinaryOp";
-  op: BuiltinBinop;
+  op: string;
```
jared-hughes, line 275: Why allow an arbitrary string here? I find it's better to know that you're limited to a certain set, even if every language doesn't use that whole set

src/IR/traverse.ts
```diff
@@ -3,12 +3,15 @@ import { IR } from ".";
 export class Path<N extends IR.Node = IR.Node> {
   _removed = false;
   visitState: VisitState | null = null;
+  root: Path;
```
jared-hughes, line 6: Would root be a `Path<Program>` always?

src/IR/IR.ts
MichalMarsalek: I did it to make BinaryOp -> MutatingBinaryOp work, since BinaryOp.op was arbitrary string. I agree but those two nodes need to match. The plugins should not rename the ops, emitters can handle that.

src/IR/traverse.ts
MichalMarsalek: It should, yes.

PR #12, opened 2022-08-23
Refactor IR

jared-hughes:
This started as removing applications and evolved to refactoring the IR completely (split `IR.ts` into files, move unrelated files into `common/`).

Some applications have not been converted yet, so I have marked this as draft. In particular:

- print + println should be their own Print node with a flag for trailingNewline
- str_length ?
- cardinality ? maybe also rename into two nodes: one for List, one for Table
- int_to_str should be unary op
- str_to_int should be unary op
- sorted should be an unary op?
- str_get_char should be its own node by analogy with ArrayGet
- contains_key, contains_value, and indexof should be their own nodes?


MichalMarsalek:
I finished the Application removal.

Since handling ops transforms will we so common, I added `Language.opMap`. It defines transforms of `BinaryOp` & `UnaryOp` to other nodes. Initially I thought it would handle op renaming too, but I cannot assign arbitrary strings to `BinaryOp.op` so the renaming stayed in the Lua emitter.

PR #13, opened 2022-08-24
implement node type annotation

MichalMarsalek:
Added `valueType` key to nodes. Builder functions now set the valueType based on child nodes. There's no type checking.
Identifiers don't know their types.

Now that I thing about it it might be better to create all the nodes as typeless and then run a plugin that would annotate the nodes (knowing the `varDeclaration`s information).


jared-hughes (commented):
To avoid duplication, it would be better to use `interface BaseNode { valueType: ValueType }`, then use `extends BaseNode` on all of the node interfaces

jared-hughes:
I agree about annotating the nodes afterwards in a type inference step. Though we also have to consider that we do a bunch of transformations on the nodes, and each will have to keep track of the types, so there should still be forms of builders that take a type.

It takes work to ensure the invariant that types are always correct, and there's duplication between getting the initial types and maintaining types later on. An alternative is to keep track of types in a [query-based](https://ollef.github.io/blog/posts/query-based-compilers.html) pattern:

- each node is given an ID
- whenever a type is needed, the full computation needs to occur, but
- we cache the type of nodes based on ID

If nodes' types always stay consistent based on ID, then this sounds pretty simple. (If the type could change without changing the ID, then we have to keep track of dependencies of nodes and invalidate the cache, but we shouldn't have this case: just make a new node with a new ID)

I'm leaning towards preferring the query-based types because it seems simpler.

Implementation note: instead of IDs, we could also just use the object references themselves as keys in a `WeakMap`. But then they won't cache across `structuredClone`s. Not sure if this is an issue.

jared-hughes (approved):

PR #14, opened 2022-09-21
implement integer subdomains

MichalMarsalek:


src/IR/types.ts
```diff
@@ -54,3 +57,27 @@ export function arrayType(
...
+  high?: bigint | number
+): ValueType {
+  return {
+    type: "integer",
+    low: typeof low === "number" ? BigInt(low) : low,
+    high: typeof high === "number" ? BigInt(high) : high,
+  };
+}
+
+export function integerTypeFromMultipleBounds(bounds: bigint[]) {
```
jared-hughes, line 72: This might be better named as a union of bounds since we might later need an intersection of bounds

jared-hughes (approved):

PR #19, opened 2022-10-16
Add Polygolf DSL

jared-hughes:
It's in a bit of a rough state, so marking this as draft for now. Pretty much everything is an s-expression.

In particular, I want to avoid the need for parentheses around statements. Requiring a semicolon would work I guess.


MichalMarsalek:
> I want to avoid the need for parentheses around statements. Requiring a semicolon would work I guess.

It should work for any node list:
`assign a (list [1; add x 1; mul y 3])`

src/frontend/lexer.ts
```diff
@@ -0,0 +1,35 @@
...
+  string: /"(?:\\.|[^"])*"/,
+  variable: /\$\w*/,
+  builtin: /\w+/,
+  lparen: "(",
+  rparen: ")",
+  lbracket: "[",
+  rbracket: "]",
+  lbrace: "{",
+  rbrace: "}",
+  star: "*",
```
MichalMarsalek, line 14: use `|` instead of `*`?

src/frontend/lexer.ts
```diff
@@ -0,0 +1,35 @@
+import moo from "moo";
+
+const tokenTable = {
+  integer: /[0-9]+/,
```
MichalMarsalek, line 4: `/-?[0-9]+/`

PR #20, opened 2022-10-22
Lang parser improvements

MichalMarsalek:
- line comments using `#`
- node annotations
- nullary nodes not requiring parens around them
- makes for range step optional
- adds symbol aliases to these common ops: add, sub, neg, mul, exp, lt, leq, eq, neq, gt, str_concat
- allows these associative ops to be written as n-ary: add, mul, str_concat
- add if-else, while


jared-hughes (approved):
Looks good to me.

Except the concern with Text vs String, but that feels like bikeshedding.

PR #23, opened 2022-10-30
Types pass

MichalMarsalek:
* [x] Set variable type based on the first assignment.
* [x] Throw if variable is not first used in an assignment.
* [x] Run types pass after variants are expanded, so that they don't interfere with each other. 
* [x] Get rid of Statement node (use Expr instead).
* [x] Let Text type have a max length parameter.
* [x] Type check all ops. 
* [x] Rename ops to snake_case.
* [x] Store pointers to source code on Nodes for error messages to use.


src/IR/types.ts
```diff
@@ -6,56 +6,89 @@ export interface IntegerType {
   low?: bigint;
   high?: bigint;
 }
+export interface TextType {
+  type: "text";
+  capacity?: number;
```
jared-hughes, line 11: `Infinity` could make more sense here instead of `undefined` to convey "unbounded." Also avoids extra checks in the operations involving it.

Alternatively, this size could be an IntegerType with the existing notations of unbounded intervals, but there's no need for the minimum length.

src/IR/types.ts
```diff
@@ -6,56 +6,89 @@ export interface IntegerType {
...
+export function setType(member: ValueType | "void" | "boolean"): ValueType {
   return {
     type: "Set",
-    member: typeof member === "string" ? simpleType(member) : member,
+    member:
+      member === "boolean"
+        ? booleanType
+        : member === "void"
+        ? voidType
+        : member,
```
jared-hughes, line 64: These five lines are repeated several times. Can benefit from a helper function.

src/IR/types.ts
jared-hughes: Or capacity could be a bigint to keep with our convention of internals using bigints.

src/common/getType.ts
```diff
@@ -23,202 +37,474 @@ export function getType(expr: Expr, program: Program): ValueType {
...
+      expectType(textType(), textType(), textType());
+      return textType(); // TODO narrow this
+    case "text_get_slice": {
+      expectType(textType(), integerType(0), integerType(0));
+      const [t, i1, i2] = types as [TextType, IntegerType, IntegerType];
+      return textType(
+        t.capacity === undefined && i2.high === undefined
+          ? undefined
+          : t.capacity === undefined
+          ? Number(i2.high! - i1.low!)
```
jared-hughes, line 472: Where does `i1.low` get checked to let you conclude it's a number?

src/languages/lua/lua.test.ts
```diff
@@ -8,87 +8,85 @@ import {
...
+          assignment("B", polygolfOp("false")),
+          assignment("a", arrayConstructor([stringLiteral("xy")])),
+          polygolfOp(
+            op,
+            ...args.map((x) => (typeof x === "string" ? id(x) : x))
+          ),
+        ])
+      ),
+      `i=0
+I=4
```
jared-hughes, line 47: Smart way to bring the types into the tests without complicating the describe blocks.

src/languages/lua/lua.test.ts
```diff
@@ -8,87 +8,85 @@ import {
...
-  testStatement(
-    "method call on string",
-    polygolfOp("str_length", stringLiteral("abc")),
-    `("abc"):len()`
-  );
-  testStatement(
-    "method call on ArrayGet",
-    polygolfOp("str_length", polygolfOp("array_get", id("A"), id("i"))),
-    `A[i+1]:len()`
-  );
```
jared-hughes, line 90: What happened to this test?

jared-hughes (commented):
Mostly just nitpicks here. Otherwise looks good.

src/IR/types.ts
MichalMarsalek: > Or capacity could be a bigint to keep with our convention of internals using bigints.

Thought about that, but array length is not `bigint` either.

src/common/getType.ts
MichalMarsalek: It doesn't thank you for being thorough with the reviews. :)

PR #26, opened 2022-11-10
add function node and valueType

MichalMarsalek:
Usage: `$increment <- (func $x:-oo..oo ($x+1));`
Recurisve functions should work but need to be type annotated:

```
$fac:(Func 0..oo 0..oo) <- (func $x:0..oo
    (conditional ($x == 0) 1 (($fac ($x-1):0..oo) * $x))
)
```

### DIFF
--- a/src/IR/exprs.ts
+++ b/src/IR/exprs.ts
@@ -96,7 +96,7 @@ export interface UnaryOp extends BaseExpr {
 /**
  * Conditional ternary operator.
  *
- * Python: [alternate,consequent][condition].
+ * Python: [alternate,consequent][condition] or consequent if condition else alternate
  * C: condition?consequent:alternate.
  */
 export interface ConditionalOp extends BaseExpr {
@@ -104,6 +104,7 @@ export interface ConditionalOp extends BaseExpr {
   condition: Expr;
   consequent: Expr;
   alternate: Expr;
+  isSafe: boolean; //whether both branches can be safely evaluated (allows for more golfing)
 }
 
 export interface Function extends BaseExpr {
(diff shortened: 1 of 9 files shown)


MichalMarsalek:
forgot to type check applications of user defined functions

src/IR/exprs.ts
```diff
@@ -96,14 +96,21 @@ export interface UnaryOp extends BaseExpr {
...
- * Python: [alternate,consequent][condition].
+ * Python: [alternate,consequent][condition] or consequent if condition else alternate
  * C: condition?consequent:alternate.
  */
 export interface ConditionalOp extends BaseExpr {
   kind: "ConditionalOp";
   condition: Expr;
   consequent: Expr;
   alternate: Expr;
+  isSafe: boolean; // whether both branches can be safely evaluated (allows for more golfing)
```
jared-hughes, line 107: `isPure` makes more sense to me

src/IR/types.ts
```diff
@@ -173,7 +191,12 @@ export function toString(a: Type): string {
 
 export function union(a: Type, b: Type): Type {
   try {
-    if (a.kind === "List" && b.kind === "List") {
+    if (a.kind === "Function" && b.kind === "Function") {
+      return functionType(
+        a.arguments.map((t, i) => union(t, b.arguments[i])),
```
jared-hughes, line 196: Function types should be contravariant in the parameter types, so this should be `intersect`, not `union`. But it's more work to implement `intersect`, so you might want to just throw an error if the parameter types differ.

jared-hughes (changes requested):
Some nitpicks and a concern: contravariance of function parameters.

PR #30, opened 2022-12-18
Traverse improvements

MichalMarsalek:
- add `skipNewNode` arg to `replaceWith` and `replaceChild`
- use the above to fix a bug in `renameIdents`
- add new function `enterProgram` to Visitor for code that should only be executed once

### DIFF
--- a/src/languages/nim/plugins.ts
+++ b/src/languages/nim/plugins.ts
@@ -36,25 +36,22 @@ const includes: [string, string[]][] = [
 
 export const addImports: Visitor = {
   name: "addImports",
-  exit(path: Path) {
-    if (path.node.kind === "Program") {
-      const program: Program = path.node;
-      const dependecies = [...program.dependencies];
-      if (dependecies.length < 1) return;
-      let imports: ImportStatement;
-      for (const include of includes) {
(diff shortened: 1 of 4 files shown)


src/languages/nim/plugins.ts
```diff
@@ -36,25 +36,22 @@ const includes: [string, string[]][] = [
...
-        }
+  enterProgram(program: Program) {
+    const dependecies = [...program.dependencies];
+    if (dependecies.length < 1) return;
+    let imports: ImportStatement;
+    for (const include of includes) {
+      if (include[0].length > dependecies.join().length - 1) break;
+      if (dependecies.every((x) => include[1].includes(x))) {
+        imports = importStatement("include", [include[0]]);
+        break;
```
jared-hughes, line 47: Small typo
```suggestion
  enterProgram(program: Program) {
    const dependencies = [...program.dependencies];
    if (dependencies.length < 1) return;
    let imports: ImportStatement;
    for (const include of includes) {
      if (include[0].length > dependencies.join().length - 1) break;
      if (dependencies.every((x) => include[1].includes(x))) {
        imports = importStatement("include", [include[0]]);
        break;
```

jared-hughes (approved):

PR #41, opened 2022-12-27
Change the way binary/unary ops are mapped

MichalMarsalek:
Changes the way unary/binary operators and their precedence is defined in Language definitions.
The previous `mapOps` signature was kinda silly and the way one needed to modify the default operator precedence was error prone.
Now, there is no default precedence, `BinaryOp` and `UnaryOp` nodes constructors now require precedence. There's a new plugin `mapPrecedenceOps` (I don't like the name very much) that takes the list of operators (or rather groups of them) in the descending order of precedence.


jared-hughes (approved):
Looks good

PR #48, opened 2022-12-29
Readonly IR

jared-hughes:
Despite the branch name, the primary contribution of this PR is a read-only IR state. This removes the need for `structuredClone`s, which are pretty expensive compared to the alternative (just pass stuff around, knowing it will not get mutated).

With the read-only IR state, I rewrote all of the plugins to use the new read-only APIs and restricted what plugins can do: each plugin is now a visitor that takes a node (and its parents) and returns `undefined` for no change or the new node if it suggests a change. Plugins are utilized in two ways:

1. `applyAll` in applyLanguage.ts: apply the plugin over the whole tree, and replace every node it suggests to replace. This mode is used in emit plugins to ensure that all of their changes are applied, such as converting all exclusive for-ranges to inclusive for-ranges.
2. one replacement at a time: this is used in the BFS search. The search creates a new program by replacing the node, then considers that new program as a valid state to push to the queue. So if a plugin suggests a replacement for each of 5 nodes, the search continues on those 5 new programs.

The key to immutable actions is `Spine`s as a replacement for the old `Path`. A spine is just a node with its parents, together with some helper functions. Spines make immutable actions quite natural, similar to how git branches make immutable commits natural as well. (I think I got the inspiration for Spines from some Anders Hejlsburg interview where he seemed to indicate it's what TypeScript uses (I am not sure about this claim)).

The main reason immutable nodes are faster than structured clones is the ability to re-use nodes: In the tree `(add A B)`, swapping the arguments to get `(add B A)` costs as little as just constructing a new "add" node and re-using the old pointers to the `A` and `B` subtrees. To get a new program with `N` changed to node `M`, it is enough to construct a new node like `N.parent` but with `N` changed to `M`, a new node like `N.parent.parent` but with `N.parent` changed to the new `N.parent`, etc. So it just needs to re-make the parent nodes to create an entirely new program tree, so replacing a node immutably is average-case O(log(N)) in the number of tree nodes.

TODO:
- [x] add a description of spines and stuff to architecture.md
- [x] consider what happens if an alternative includes another, like if swapBinaryOps was used (for some reason) as an emitPlugin for `a + (b + c))`. This has realistic potential in nested exclusive for-loops converted to inclusive for-loops. I'm not sure if I got the visit order right here.
- [x] comment the code a bit more


src/common/expandVariants.ts
```diff
@@ -1,108 +1,42 @@
...
  */
 export function expandVariants(program: IR.Program): IR.Program[] {
-  const structure: Variant = getVariantsStructure(program);
-  const expansionCount = countVariantExpansions(structure);
-  if (expansionCount > 1000) {
-    throw new Error(`Too many variants (${expansionCount}).`);
+  return allVariantOptions(program) as IR.Program[];
+}
+
+function allVariantOptions(node: IR.Node): IR.Node[] {
```
MichalMarsalek, line 17: Previously, the total amount of expanded variants was calculated before the actual expansion happened so that the expansion is not performed when it was too high. Creating more variants than what can be processed in a reasonable time (or even fit in memory) probably cannot happen by accident, but still, I think it was a nice feature to have.

src/plugins/idents.ts
```diff
@@ -85,41 +85,42 @@ function defaultShouldAlias(name: string, freq: number): boolean {
...
+      })) {
+        const freq = timesUsed.get(name) ?? 0;
+        timesUsed.set(name, freq + 1);
       }
+      // apply
+      const assignments: (IR.Assignment & { variable: Identifier })[] = [];
+      const replacedDeep = spine.withReplacer((node) => {
+        if (
+          node.kind === "Identifier" &&
+          node.builtin &&
```
MichalMarsalek, line 106: Should this be `!node.builtin`?

src/plugins/divisionOps.ts
```diff
@@ -1,20 +1,20 @@
...
-    const program = path.root.node;
+  visit(node, spine) {
+    const program = spine.root.node;
     if (node.kind === "PolygolfOp" && node.op === "mod") {
       const rightType = getType(node.args[1], program);
       if (rightType.kind !== "integer")
         throw new Error(`Unexpected type ${JSON.stringify(rightType)}.`);
       if (leq(0n, rightType.low)) {
-        node.op = "rem";
+        return copyType(node, polygolfOp("rem", ...node.args));
```
MichalMarsalek, line 14: Should copying type not be done implicitly?

MichalMarsalek (commented):
Good job! Noticed just a couple of things.

src/common/debug.ts
```diff
@@ -1,13 +0,0 @@
-import { Program } from "../IR";
```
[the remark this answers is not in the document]
jared-hughes: It wasn't used anywhere, so I deleted it instead of changing to the new `applyLanguage` signature. Should I restore it?

src/plugins/idents.ts
jared-hughes: This is `aliasBuiltins`, so the `node.builtin` check is to ensure it's only aliasing builtins. I guess that's because non-builtins (essentially) always have a one-letter name.

PR #49, opened 2022-12-29
Markdown tests

MichalMarsalek:
Adds support for markdown defined test suites. The markdown tests are automagically compiled to typescript Jest code.

To create such suite, create a markdown file with a ".test.md" extension.
Headings are used to group the tests (using Jest's `describe`).
Code blocks are used to define the actual tests.
All other markdown syntax is ignored.

Each codeblock is required to have the language annotation and optionally other test parameters/options/commands (all space-separated).
If the language is `polygolf` (with no other options) it denotes an input, otherwise an output.
Multiple output blocks can relate to a single input block.
If a codeblock title contains "skip" it's not used to generate tests.

See loops.test.md for how it is used for plugin testing.
See nim.test.md for how it is used for language testing.

Implements #59 .


src/markdown-tests/build.ts
```diff
@@ -0,0 +1,217 @@
...
+}
+
+/** Compiles a single markdown test suite source to a Jest source. */
+function compileSuite(markdown: string): string {
+  const suite = parseSuite(markdown);
+  if (suite.children.length === 0) return "";
+  return emitSuite(suite);
+}
+
+function readdirSync(p: string, a: string[] = []) {
```
jared-hughes, line 202: Might need a more descriptive name. `readdirSyncRecursive` maybe? or `readdirSyncDeep`.

src/markdown-tests/build.ts
```diff
@@ -0,0 +1,217 @@
...
+  if (fs.statSync(p).isDirectory())
+    fs.readdirSync(p).map((f) =>
+      readdirSync(a[a.push(path.join(p, f)) - 1], a)
+    );
+  return a;
+}
+
+for (const file of readdirSync(process.cwd())) {
```
jared-hughes, line 210: An approach that avoids `node_modules` and works regardless of the path `npm run test:build` is called from. Seems janky though since it depends on the output directory of the node script. Maybe just filter out `node_modules` in `readdirSync`?
```suggestion
for (const file of readdirSync(path.join(__dirname, "../../src"))) {
```
Though idk why I want to filter out node_modules. Some performance instinct ig, but we have no bottleneck here.

src/markdown-tests/build.ts
```diff
@@ -0,0 +1,217 @@
...
+      });
+    }
+  }
+  return result;
+}
+
+/** Emits a single markdown test. */
+function emitSuite(describe: Describe): string {
+  return `import parse from "frontend/parse";
+import applyLanguage from "@/common/applyLanguage";
```
jared-hughes, line 129: Might want a comment here to the effect of: the `@/` path is defined relatively in tsconfig.json, pointing to the `src` directory

package.json
```diff
@@ -15,8 +15,9 @@
     "test:formatting": "prettier-check .",
     "test:lint": "eslint \"src/**\"",
     "test:jest": "jest --config jest.config.js",
-    "test": "npm run test:formatting && npm run test:typecheck && npm run test:lint && npm run test:jest",
-    "cli": "npm run build && node --enable-source-maps dist/cli.js"
+    "test": "npm run test:formatting && npm run test:typecheck && npm run test:lint && npm run test:build && npm run test:jest",
+    "cli": "npm run build && node --enable-source-maps dist/cli.js",
+    "test:build": "node \"dist/markdown-tests/build.js\""
```
jared-hughes, line 20: Ensure the markdown tests are built before testing.
```suggestion
    "test:build": "npm run build && node \"dist/markdown-tests/build.js\""
```

jared-hughes (changes requested):

src/markdown-tests/build.ts
```diff
@@ -0,0 +1,217 @@
...
+${emitDescribe(describe)}`;
+}
+
+function emitNode(node: Describe | Test): string {
+  if ("children" in node) return emitDescribe(node);
+  return emitTest(node);
+}
+
+function stringify(x: string): string {
+  return emitStringLiteral(x, [
```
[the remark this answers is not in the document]
MichalMarsalek: I had `JSON.stringify` in the original commit but I changed it because I didn't like how it had to escape all the `"`. I wasn't trying to golf the tests but I was trying to make then a little more readable.

src/markdown-tests/build.ts
MichalMarsalek: That came to my mind but I thought depending on the path it is called from makes sense?

src/markdown-tests/build.ts
MichalMarsalek: Is this not a common thing to have?

jared-hughes: I spent some time googling TypeScript `@/`, thinking it was a language feature. Would help clarity for future readers

src/common/applyLanguage.ts
```diff
@@ -4,13 +4,39 @@ import { Language, defaultDetokenizer, Plugin } from "./Language";
...
+  objectiveFunction?: (x: string) => number
+): SearchOptions {
+  return {
+    level,
+    objective,
+    objectiveFunction:
+      objectiveFunction ??
+      (objective === "bytes"
+        ? (x) => Buffer.byteLength(x, "utf-8")
+        : (x) => x.length),
```
jared-hughes, line 27: Note `"😊".length` gives 2, but we expect 1. Try copying [charLen](https://github.com/code-golf/code-golf/blob/13733cfd472011217031fb9e733ae9ac177b234b/js/_util.ts#L7) from code-golf/code-golf.

src/common/applyLanguage.ts
```diff
@@ -4,13 +4,39 @@ import { Language, defaultDetokenizer, Plugin } from "./Language";
 import { programToSpine } from "./Spine";
 import polygolfLanguage from "../languages/polygolf";
 
+export type OptimisationLevel = "none" | "heuristic" | "full";
```
jared-hughes, line 7: "heuristic" doesn't seem to be handled. Put in TODO comment, or maybe this is more appropriate as a GH issue.

jared-hughes (changes requested):
Only blocking change is the chars scoring (JS string length measures UTF-16 surrogates, not 
 characters). This could also use a test.

PR #53, opened 2022-12-30
Add Swift

MeWhenI:
Currently the way some ops with odd syntax are added is a bit janky, this is fixable using f-strings to define their emit (see: text_to_int, text_len, print, repeat).


MichalMarsalek:
This needs main merged to work with the readonly IR and also some tests.

src/languages/swift/index.ts
```diff
@@ -0,0 +1,149 @@
...
+            "String"
+          ),
+      ],
+      [
+        "byte_to_char",
+        (x) => functionCall([functionCall([x[0]], "UnicodeScalar")], "String"),
+      ],
+      ["text_length", (x) => methodCall(x[0], [], "count")],
+      // ["text_length_chars", (x) => methodCall(x[0], [], "count")],
+      // ["text_length_bytes", (x) => methodCall(methodCall(x[0], [], "utf8"), [], "count")]
```
jared-hughes, line 73: `text_length_chars` and `text_length_chars`: intended to be missing?

jared-hughes (approved):

src/languages/swift/index.ts
MeWhenI: Yes, I preemptively wrote in behavior for ops that should be added alongside #71

PR #58, opened 2022-12-31
Add GolfScript

MeWhenI:
Not gonna be winning any GolfScript golds at this stage, buts its a fully functional MVP. ⛳


MichalMarsalek:
This needs main merged to work with the readonly IR and also some tests.

src/languages/golfscript/index.ts
```diff
@@ -0,0 +1,92 @@
...
+  extension: "txt", // Golfscript doesn't appear to have a custom extension
+  emitter: emitProgram,
+  golfPlugins: [flipBinaryOps, evalStaticExpr, golfLastPrint()],
+  emitPlugins: [useIndexCalls()],
+  finalEmitPlugins: [
+    mapOps([
+      ["true", (_) => id("1", true)],
+      ["false", (_) => id("0", true)],
+      ["println", (x) => functionCall(x, "puts")],
+      ["print", (x) => functionCall(x, "print")],
```
jared-hughes, line 28: Does the current system make it feasible to use `p` for printing integers?

src/languages/golfscript/index.ts
```diff
@@ -0,0 +1,92 @@
...
+      ["neq", "=!"],
+      ["geq", "(>"],
+      ["join", "''*"],
+      ["text_reversed", "-1%"],
+      ["text_get_char", "=[]+''+"],
+      ["byte_to_char", "[]+''+"],
+      ["max", "[]++$1="],
+      ["min", "[]++$0="],
+
+      ["argv_get", "a\\="],
```
jared-hughes, line 78: Is the swap `\` necessary? In testing, it seemed to work with no `\`.

jared-hughes (approved):

src/languages/golfscript/index.ts
MeWhenI: No, good point

src/languages/golfscript/index.ts
MeWhenI: It does, yes. In the current system, `println` and `print` can be feasibly implemented as "n" and "", though this likely isn't a very future-proofed approach.

src/languages/golfscript/index.ts
MeWhenI: Fixed

PR #60, opened 2023-01-02
Allow emitters to return trees of tokens instead of arrays

MichalMarsalek:
What this means in practice is there's no need for intermediate spreading. Flattening to array of tokens happens at the very end.


src/common/Language.ts
```diff
@@ -36,7 +36,9 @@ export interface Plugin {
   allOrNothing?: boolean;
 }
 
-export type Detokenizer = (tokens: string[]) => string;
+interface TokenTreeArray extends Array<string | TokenTreeArray> {}
```
jared-hughes, line 39:
```suggestion
type TokenTreeArray = Array<string | TokenTreeArray>;
```

src/common/Language.ts
```diff
@@ -59,7 +61,9 @@ export function defaultDetokenizer(
   whitespace: WhitespaceInsertLogic = defaultWhitespaceInsertLogic,
   indent = 1
 ): Detokenizer {
-  return function (tokens: string[]): string {
+  return function (tokenTree: TokenTree): string {
+    // @ts-expect-error
+    const tokens: string[] = [tokenTree].flat(Infinity); // it seems ts doesn't understand flattening the tree
```
jared-hughes, line 66: TS could be convinced to like this by doing some wrapper function + recursion, but this solution is elegant enough that it warrants the `@ts-expect-error`

jared-hughes (approved):

PR #64, opened 2023-01-03
Packing

### LINKED ISSUE #62: Automatically pack code with another option
It would be nice if there was a `--chars` option or something that tries packing the code, and if it is shorter, then outputs the packed code instead.

### LINKED ISSUE #63: Allow compiling to all languages at once
So do like `polygolf -i src/programs/fibonacci.polygolf -l all -o fibonacci` and it will output to `fibonacci.lua`, `fibonacci.nim`, etc.

Another possible thing is that instead of having `-o fibonacci`, you can have an option like `--cg` that will automatically upload everything to code.golf (of course, you have to input your token or whatever).

MichalMarsalek:
* Closes #62
* Closes #63 
* Implements two printers for hardcoded outputs: constans holes and (low numbers) sequence holes


src/cli.ts
```diff
@@ -35,38 +44,50 @@ const options = yargs()
...
+    const result = applyLanguage(
+      lang,
+      prog,
+      searchOptions("full", options.chars === true ? "chars" : "bytes")
+    );
+    if (options.output !== undefined) {
+      fs.mkdirSync(path.dirname(options.output), { recursive: true });
+      fs.writeFileSync(options.output + "." + lang.extension, result);
+    } else {
+      console.log(result);
```
MeWhenI, line 70: The output generated on the command line when compiling `polygolf -i XXX.polygolf -l all` could benefit from spacing between and/or labeling of the various solutions it outputs.

PR #65, opened 2023-01-04
Assembly backend

CatsAreFluffy:
Prototype assembly backend. Solutions are generally around twice as long as hand-optimized assembly solutions. Needs updating.


MichalMarsalek:
If this was merged with all the comments and TODOs, I wouldn't be able to deal with them later.

MichalMarsalek:
This also needs main merged to work with the readonly IR and also some tests.

PR #68, opened 2023-01-11
typecheck before golfing

### LINKED ISSUE #67: When applying an operation to a variable in a loop, Polygolf tries to replace it with a pre-calculated expression
For example this code:
```
$s <- 5;
println $s;
$s <- ($s * $s);
println $s;
```
Compiles correctly into something like this:
```lua
s=5
print(s)
s=25
print(s)
```

But it still does this even if we do it in a loop:
```
$s <- 5;

for $i 0 5 {
  $s <- ($s * $s);
  println $s;
};
```
=>
```lua
s=5
for i=0,4 do s=25
print(s)end
```

MichalMarsalek:
Typechecks input programs (each included variant) before doing anything else.

Fixes #67

### DIFF
--- a/src/common/applyLanguage.ts
+++ b/src/common/applyLanguage.ts
@@ -197,7 +197,8 @@ function golfProgram(
   return shortestSoFar;
 }
 
-/** Typecheck a program by asking all nodes about their types. */
+/** Typecheck a program by asking all nodes about their types.
+ * Throws an error on a type error; otherwise is a no-op. */
 function typecheck(program: Program) {
   const spine = programToSpine(program);
   spine.everyNode((x) => {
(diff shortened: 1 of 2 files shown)


src/common/applyLanguage.ts
```diff
@@ -120,13 +126,15 @@ function golfProgram(
   program: IR.Program,
   golfPlugins: Plugin[],
   finalEmit: (ir: IR.Program) => string,
-  objective: (x: string) => number
+  objective: (x: string) => number,
+  skipTypecheck = true
```
jared-hughes, line 130: This is default `true`, but the others are default `false`. Intended?

src/common/applyLanguage.ts
```diff
@@ -188,3 +196,12 @@ function golfProgram(
   }
   return shortestSoFar;
 }
+
+/** Typecheck a program by asking all nodes about their types. */
```
jared-hughes, line 200:
```suggestion
/** Typecheck a program by asking all nodes about their types.
 * Throws an error on a type error; otherwise is a no-op. */
```

jared-hughes (approved):

PR #70, opened 2023-01-22
Hashing

MichalMarsalek:
Introduces plugin which replaces `tableLiteral[key]` with `listLiteral[hash(key)%mod%len(listLiteral)]`.

### DIFF
--- a/src/plugins/hashing.ts
+++ b/src/plugins/hashing.ts
@@ -1,0 +1,66 @@
+import { getType } from "@/common/getType";
+import { Plugin } from "@/common/Language";
+import {
+  Expr,
+  functionCall,
+  id,
+  int,
+  listConstructor,
+  polygolfOp,
+  StringLiteral,
+} from "@/IR";
+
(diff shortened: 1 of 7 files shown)


src/plugins/hashing.ts
```diff
@@ -0,0 +1,121 @@
...
+  maxMod: number
+): [(Expr | undefined)[], number] {
+  let width = table.length;
+  const hashedTable: [number, Expr][] = table.map((x) => [
+    hashFunc(x[0]),
+    x[1],
+  ]);
+  while (true) {
+    for (let mod = width; mod <= maxMod; mod++) {
+      const result: (Expr | undefined)[] = Array(width).fill(undefined);
```
jared-hughes, line 89: Assuming this is a bottle neck here: use a `Set` to check for collisions, then only populate an array once you find a valid modulus.

jared-hughes (approved):
Looks good. I left one suggestion for performance.

PR #71, opened 2023-01-26
Text subtypes

### LINKED ISSUE #51: Distinguish treatment of bytes and characters
There is not a default behavior for `text-length`; replacing it with 2 new ops, `text-length-bytes` and `text-length-chars` reduces ambiguity. Same situation for `text-get-slice`

MichalMarsalek:
This PR mainly introduces new Text subtypes. Also, closes #51.

## Backend
The subtypes are modeled by redefining `TextType` as
```typescript
export interface TextType {
  readonly kind: "text";
  readonly codepointLength: IntegerType;
  readonly isAscii: boolean;
}
```

- Motivation for adding `.isAscii`: Knowing the text contains only ascii characters is useful, because it allows golfing plugins which freely swap the text ops that are bytes vs char (codepoint) based.
- Motivation for adding lower bound for length: This allows to represent text type of values that have length exactly 1.
- Motivation for expressing the length in terms of codepoints rather than in terms of bytes: This allows to express both a single byte text and single codepoint text.

Note that this gives us a "character" type (or two actually) but it is a subtype of text (so like in Python or Javascript, there's no distinct character type).

## Front end
On the front end, we have

- `Text` - an unbounded text
- `Ascii` - an unbounded text with Ascii characters only
- `(Text 10..20)` - text containing 10 to 20 codepoints
- `(Ascii 10..20)` - text containing 10 to 20 ascii characters

For backwards compatibility, we also have

- `(Text 5)` as an alias to `(Text 0..5)`

And for consistency

- `(Ascii 5)` as an alias to `(Ascii 0..5)`

## New/renamed opcodes
Several opcodes were added or renamed so that there are both byte based and codepoint based variants for ops where it matters.

## New plugin
New plugin was introduced that take advantage of the new subtypes.


src/IR/opcodes.ts
```diff
@@ -33,20 +33,23 @@ export const BinaryOpCodeArray = [
...
   "table_get",
-  "text_get_byte",
   // other
   "list_push",
   "text_concat",
   "repeat",
   "text_contains",
-  "text_find",
+  "text_codepoint_find", // (text_codepoint_find a b) returns the codepoint-0-index of the start of the first occurence of b in a or -1 if it is not found
+  "text_byte_find", // (text_codepoint_find a b) returns the byte-0-index of the start of the first occurence of b in a or -1 if it is not found
```
jared-hughes, line 42:
```suggestion
  "text_byte_find", // (text_byte_find a b) returns the byte-0-index of the start of the first occurence of b in a or -1 if it is not found
```

src/common/getType.ts
```diff
@@ -400,57 +404,82 @@ function getOpCodeType(
...
+        t.isAscii
+      );
     }
     case "text_contains":
       expectType(textType(), textType());
       return booleanType;
-    case "text_find":
+    case "text_codepoint_find":
+    case "text_byte_find":
       expectType(textType(), textType());
```
jared-hughes, line 425: What happens if the second string is length-0?

src/common/getType.ts
```diff
@@ -400,57 +404,82 @@ function getOpCodeType(
...
-      return integerType(-1, (types[0] as TextType).capacity - 1);
+      return integerType(
+        -1,
+        mul(
+          sub((types[0] as TextType).codepointLength.high, 1n),
```
jared-hughes, line 429:
```suggestion
          sub((types[0] as TextType).codepointLength.high, (types[1] as TextType).codepointLength.low),
```

Slightly narrower, probably doesn't matter.

src/common/getType.ts
```diff
@@ -400,57 +404,82 @@ function getOpCodeType(
...
+      );
     case "text_split":
       expectType(textType(), textType());
       return listType(types[0]);
-    case "text_get_char":
+    case "text_get_byte":
+    case "text_get_codepoint":
       expectType(textType(), integerType(0));
-      return textType(1);
+      return textType(integerType(1, 1), (types[0] as TextType).isAscii);
```
jared-hughes, line 441: This suggests the potential of keeping track of byte length separate from codepoint length: `text_get_byte` would alway returns a 1-byte string, while `text_get_codepoint` gives a 1-codepoint string. Not in the scope of this PR, maybe just something to keep in mind.

src/common/getType.ts
```diff
@@ -485,42 +514,58 @@ function getOpCodeType(
...
-        10n ** BigInt(t.capacity) - 1n
+        1n - 10n ** (t.codepointLength.high - 1n),
+        10n ** t.codepointLength.high - 1n
       );
     }
     case "bool_to_int":
       expectType(booleanType);
       return integerType(0, 1);
-    case "byte_to_char":
+    case "byte_to_text":
```
jared-hughes, line 546: It's not clear what `byte_to_text` does. `byte_to_char` is a more descriptive name here — maybe `byte_to_char_ascii` to clarify encoding. Similarly, the next builtin could be `int_to_char_codepoint`.

src/common/getType.ts
```diff
@@ -485,42 +514,58 @@ function getOpCodeType(
...
       expectType(integerType(0, 255));
-      return textType(1);
+      return textType(integerType(1n, 1n), (types[0] as any).high < 128n);
+    case "int_to_codepoint":
+      expectType(integerType(0, 0x10ffff));
+      return textType(integerType(1n, 1n), (types[0] as any).high < 128n);
```
jared-hughes, line 551:
```suggestion
      return textType(integerType(1n, 1n), (types[0] as IntegerType).high < 128n);
    case "int_to_codepoint":
      expectType(integerType(0, 0x10ffff));
      return textType(integerType(1n, 1n), (types[0] as IntegerType).high < 128n);
```

src/common/getType.ts
```diff
@@ -545,22 +591,28 @@ function getOpCodeType(
       return voidType;
     case "text_replace": {
       expectType(textType(), textType(), textType());
-      const a = types[0] as TextType;
-      const c = types[2] as TextType;
-      return textType(a.capacity * c.capacity);
+      const [a, c] = [types[0], types[2]] as TextType[];
+      return textType(
+        getArithmeticType("mul", a.codepointLength, c.codepointLength),
```
jared-hughes, line 596: Could be narrowed by dividing by `b.codepointLength`, which would also check that b's length is not zero.

src/common/getType.ts
```diff
@@ -579,11 +631,9 @@ function getOpCodeType(
         ["T1", (x) => x[0]],
         ["T2", (x) => x[1]]
       )[1];
+    case null:
+      throw new Error("Cannot determine type based on null opcode.");
```
jared-hughes, line 635: Would be good to clarify (in the error message, perhaps) what is a null opcode. Does it represent that a backend transformation has created a language-specific node?

src/languages/lua/index.ts
```diff
@@ -14,6 +14,7 @@ import { tempVarToMultipleAssignment } from "../../plugins/tempVariables";
 import { evalStaticExpr } from "../../plugins/static";
 import { flipBinaryOps } from "../../plugins/binaryOps";
 import { golfLastPrint } from "../../plugins/print";
+// import { useEquivalentTextOp } from "../../plugins/textOps";
```
jared-hughes, line 17:
```suggestion
```

src/plugins/textOps.ts
```diff
@@ -0,0 +1,40 @@
...
+  visit(node, spine) {
+    if (node.kind !== "PolygolfOp") return;
+    const program = spine.root.node;
+    const typeArg0 = getType(node.args[0], program);
+    if (typeArg0.kind === "text" && typeArg0.isAscii) {
+      for (const eqPair of textOpsEquivalenceAscii) {
+        if (eqPair.includes(node.op)) {
+          return { ...node, op: eqPair[1 - eqPair.indexOf(node.op)] };
+        }
+      }
```
jared-hughes, line 30: This could really use a `bidirectionalMapGet()` helper to clarify what's going with the arrays (at first, I thought it was just an indirect implementation of a normal `Map`) and to reduce the code duplicated once more below.

src/common/getType.ts
MichalMarsalek: > Could be narrowed by dividing by `b.codepointLength`, which would also check that b's length is not zero.

`text_replace "xyz" "abc" "X";` is `"xyz":(Text 3..3)` but dividing would suggest `(Text 1..1)`. Unless I missunderstood what you meant?

src/common/getType.ts
MichalMarsalek: A wanted to avoid the slighly ambiguous word character. The output might not be ascii since the input can be > 127. Maybe `byte_to_text_byte`? Not sure what's best here, however, for the second one, `int_to_codepoint` seems very clear to me.

jared-hughes: I was bikeshedding. `byte_to_text` is good enough. And `int_to_codepoint` is plenty descriptive.

src/common/getType.ts
jared-hughes: You're right

PR #73, opened 2023-02-05
Fix #38

### LINKED ISSUE #38: Variants parsing ambiguity
Currently
```
{
 {
  $a <- 0;
 }
}
```
throws `Ambiguous Parse of Code.`. But it should be parsed unambiguously.

MichalMarsalek:
FIxes #38 .

It's a bit ugly, I'll be glad to see how to write it better.

### DIFF
--- a/src/frontend/grammar.ne
+++ b/src/frontend/grammar.ne
@@ -25,9 +25,8 @@ import {
 main -> (sexpr_stmt | variants):+ {% d => refSource(program(blockOrSingle(d[0].map(id))), d[0][0]) %}
 
 variant ->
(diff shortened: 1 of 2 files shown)


src/frontend/grammar.ne
```diff
@@ -25,7 +25,8 @@ import {
 main -> (sexpr_stmt | variants):+ {% d => refSource(program(blockOrSingle(d[0].map(id))), d[0][0]) %}
 
 variant ->
-  (sexpr_stmt | variants):* {% d => refSource(blockOrSingle(d[0].map(id)), d[0].length > 0 ? d[0][0] : undefined) %}
```
MichalMarsalek, line 28: I couldn't remember why the condition `d[0].length > 0` was there so I removed it.

PR #79, opened 2023-02-08
Fix #78

### LINKED ISSUE #78: Invalid multivar assignment in Nim
Compiling
```polygolf
if (1 == 1) {
 $a:0..oo <- 1;
 $b:0..oo <- 2;
 $c:0..oo <- ($a + $b);
};
``` 
to Nim produces
```Nim
if 1==1:
 var(a,b,c)=(1,2,a+b)
```
which throws `Error: undeclared identifier: 'a'`. A non-erroring code it could produce is
```Nim
if 1==1:
 var
  a=1
  b=2
  c=a+b
```
Note: the `if` statement is just there to force the variable declaration to be indented, causing the erroneous version of the assignment to be used.

### LINKED ISSUE #104: Combine variable inits
Add plugin that golfs

```python
x=0
y=0
```

to

```python
x=y=0
```

### LINKED ISSUE #162: nim doesn't switch to indentation/parentheses with `var`
**Describe the bug**
Using `var` in a loop still uses semicolons instead of parens/indentation.

**To Reproduce**
The following code:
```
for $i 0 5 {
  $s <- $i;
  println $s;
};
```
compiles to (Nim):
```nim
for i in..4:var s=i;s.echo
```

**Expected behavior**
```nim
for i in..4:(var s=i);s.echo
# OR
for i in..4:(var s=i;s.echo)
# OR
for i in..4:
 var s=i
 s.echo
```

MichalMarsalek:
Fixes #78 . Closes #104 . Fixes #162  .


src/languages/nim/emit.ts
```diff
@@ -22,51 +21,40 @@ function emitMultiExpr(expr: IR.Expr, parent: IR.Node): TokenTree {
...
-        children.map((stmt) => emitStatement(stmt, expr)),
-        "\n"
-      ),
-      "$DEDENT$",
-    ];
+  let inner = [];
+  let needsBlock = false;
+  for (const child of children) {
+    const needsNewline =
+      "consequent" in child || "children" in child || "body" in child;
```
jared-hughes, line 28: Compare with `containsMultiExpr` from common/emit.ts

```ts
export function containsMultiExpr(exprs: readonly IR.Expr[]): boolean {
  for (const expr of exprs) {
    if ("consequent" in expr || "children" in expr || "body" in expr) {
      return true;
    }
  }
  return false;
}
```

These could have a common helper function, in case we ever add a node with a new multi-expr.

src/plugins/block.ts
```diff
@@ -0,0 +1,216 @@
...
+          if (collectPredicate(expr, childSpine as Spine<Expr>, collected)) {
+            collected.push(expr as any as T);
+          } else if (collectPredicate(expr, childSpine as Spine<Expr>, [])) {
+            if (collected.length > 1) {
+              newNodes.push(...transform(collected));
+              changed = true;
+            } else {
+              newNodes.push(...collected);
+            }
+            collected = [expr as any as T];
```
jared-hughes, line 52: `as any` is unnecessary because T is a narrower type to Expr.

```suggestion
            collected.push(expr as T);
          } else if (collectPredicate(expr, childSpine as Spine<Expr>, [])) {
            if (collected.length > 1) {
              newNodes.push(...transform(collected));
              changed = true;
            } else {
              newNodes.push(...collected);
            }
            collected = [expr as T];
```

src/plugins/block.ts
```diff
@@ -0,0 +1,216 @@
...
+        } else {
+          newNodes.push(...collected);
+        }
+        if (changed) return block(newNodes);
+      }
+    },
+  };
+}
+
+const declared: Set<string> = new Set<string>();
```
jared-hughes, line 76: Slight style nitpick. But it's more of a smell that a global variable is used in a visitor. It's probably fine though. We'll always visit starting at the program node. 
```suggestion
const declared = new Set<string>();
```

jared-hughes (approved):

src/languages/nim/emit.ts
MichalMarsalek: > Compare with `containsMultiExpr` from common/emit.ts
> 
> ```ts
> export function containsMultiExpr(exprs: readonly IR.Expr[]): boolean {
>   for (const expr of exprs) {
>     if ("consequent" in expr || "children" in expr || "body" in expr) {
>       return true;
>     }
>   }
>   return false;
> }
> ```
> 
> These could have a common helper function, in case we ever add a node with a new multi-expr.

I think the problem is, that at a Polygolf level the distinction is missing and language targets can disagree on what needs a block and what doesn't.

PR #82, opened 2023-02-11
Fix #44

### LINKED ISSUE #44: Nim Spacing before $
Compiling 
```
print (int_to_text 1);
``` 
to Nim produces 
```Nim
echo$1
```
. This is a syntax error, it should produce 
```Nim
echo $1
```

### LINKED ISSUE #106: Nim - function argument starting with `(`
**Language**
Nim

**Polygolf source**

```
println ((1:0..1 + 1) * 2);
```

**Current output**

```
echo(1+1)*2
```

**Expected output**

```
echo (1+1)*2
```

MichalMarsalek:
Fixes #44 
FIxes #106
This is fixed by making the `letter` + `symbol` combination require a space by default and making the cases where the spaces should'nt be inserted explicit.

### DIFF
--- a/src/languages/nim/index.ts
+++ b/src/languages/nim/index.ts
@@ -108,18 +108,26 @@ const nimLanguage: Language = {
     renameIdents(),
     addVarDeclarations,
   ],
-  detokenizer: defaultDetokenizer(
-    (a, b) =>
-      a !== "" &&
-      b !== "" &&
-      ((/[A-Za-z0-9_]/.test(a[a.length - 1]) && /[A-Za-z0-9_]/.test(b[0])) ||
-        ("=+-*/<>@$~&%|!?^.:\\".includes(a[a.length - 1]) &&
-          "=+-*/<>@$~&%|!?^.:\\".includes(b[0])) ||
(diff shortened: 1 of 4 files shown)


src/languages/nim/index.ts
```diff
@@ -110,15 +110,26 @@ const nimLanguage: Language = {
...
-    (a, b) =>
-      a !== "" &&
-      b !== "" &&
-      ((/[A-Za-z0-9_]/.test(a[a.length - 1]) && /[A-Za-z0-9_]/.test(b[0])) ||
-        ("=+-*/<>@$~&%|!?^.:\\".includes(a[a.length - 1]) &&
-          "=+-*/<>@$~&%|!?^.:\\".includes(b[0])) ||
-        (/[A-Za-z]/.test(a[a.length - 1]) && b[0] === `"`))
-  ),
+  detokenizer: defaultDetokenizer((a, b) => {
+    if (a === "" || b === "") return false; // glue token
```
MeWhenI, line 114: Explicitly using something like "$NO_WHITESPACE$" or "$GLUE_TOKEN$" rather than an empty string may make it more clear at a glance why this is being used in the emitter.

PR #86, opened 2023-02-11
add equalityToInequality plugin

MichalMarsalek:
Golfs
```python
x%10==0
```

to

```python
x%10<1
```

### DIFF
--- a/src/languages/golfscript/index.ts
+++ b/src/languages/golfscript/index.ts
@@ -3,6 +3,7 @@ import { defaultDetokenizer, Language } from "../../common/Language";
 
 import emitProgram from "./emit";
 import {
+  equalityToInequality,
   mapOps,
   mapPrecedenceOps,
   plus1,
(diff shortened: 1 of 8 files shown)


MeWhenI:
It seems something is going wrong when compiling polygolf to other langs,
```polygolf
$a:1..10 <- 5;
$a != 1;
```
currently compiles to Python as
```python
a=5
a>2
```

MichalMarsalek:
> It seems something is going wrong when compiling polygolf to other langs,
> 
> ```
> $a:1..10 <- 5;
> $a != 1;
> ```
> 
> currently compiles to Python as
> 
> ```python
> a=5
> a>2
> ```

Hmm, yeah and my test for != didn't catch it because it's wrong...

MeWhenI:
This is likely related to how the mod operator affects int ranges but I'd say 
```polygolf
$a:4..5 <- 4;
($a mod 3) != 1;
($a mod 3) != 2;
```
ought compile to Python as 
```Python
a=4
a%3>1
a%3<2
```
rather than
```Python
a=4
a%3!=1
a%3<2
``` as it should recognize `($a mod 3)` is of type `1..2`

MichalMarsalek:
> as it should recognize `($a mod 3)` is of type `1..2`

Yes, this should be a different issue/PR.

PR #95, opened 2023-02-12
Fix Golfscript identifier bug

MeWhenI:
Prevents golfscript from naming an identifier `n`, as `n` is a built-in which will break printing behavior if modified

### DIFF
--- a/src/languages/golfscript/index.ts
+++ b/src/languages/golfscript/index.ts
@@ -78,7 +78,17 @@ const golfscriptLanguage: Language = {
       ["argv_get", "a="],
     ]),
     addImports,
-    renameIdents(),
+    renameIdents({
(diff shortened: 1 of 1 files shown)


src/languages/golfscript/index.ts
```diff
@@ -78,7 +78,17 @@ const golfscriptLanguage: Language = {
       ["argv_get", "a="],
     ]),
     addImports,
-    renameIdents(),
+    renameIdents({
+      // Custom Ident generator prevents `n` from being used as an ident, as it is predefined to newline and breaks printing if modified
+      preferred(original: string) {
+        if (/n/i.test(original[0])) return ["m", "M"];
```
MichalMarsalek, line 84:
```suggestion
        if (/n/i.test(original[0])) return ["N", "m", "M"];
```
does this work?

PR #108, opened 2023-02-13
Swift bugfixes

MeWhenI:
* Don't use variants that require bigints (Swift lacks support)
* Remove improperly implemented ops (trunc_div and rem)
* Fix whitespace behavior with custom detokenizer


src/languages/swift/index.ts
```diff
@@ -110,9 +111,7 @@ const swiftLanguage: Language = {
       [
         ["mul", "*"],
         ["div", "/"],
-        ["trunc_div", "/"],
         ["mod", "%"],
-        ["rem", "%"],
```
MichalMarsalek, line 115: Checking on code.golf, I think it's the other way around, Swift actually has `trunc_div` and `rem` and no `div` and `mod`.

MeWhenI: right, whoops

PR #109, opened 2023-02-13
Add finiteFlatten function

MeWhenI:
Using the builtin `flat` method on `tokenTree` caused TypeScript to throw `Type instantiation is excessively deep and possibly infinite.ts(2589)`. This introduces a `finiteFlatten` method which TypeScript is ok with.

### DIFF
--- a/src/common/Language.ts
+++ b/src/common/Language.ts
@@ -44,9 +44,15 @@ export type TokenTree = string | TokenTreeArray;
 export type Detokenizer = (tokens: TokenTree) => string;
 export type WhitespaceInsertLogic = (a: string, b: string) => boolean;
 
-export function finiteFlatten(tokenTree: TokenTree): string[] {
-  if (typeof tokenTree === "string") return [tokenTree];
-  return tokenTree.map(finiteFlatten).flat(1);
+export function flattenTree(tokenTree: TokenTree): string[] {
+  var flattened: string[] = [];
+
+  function stepTree(t: TokenTree) {
+    if (typeof t === "string") flattened.push(t);
+    else t.map(stepTree);
+  }
+  stepTree(tokenTree);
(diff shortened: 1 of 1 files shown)


MeWhenI:
finiteFlatten is exported so that other langs with custom detokenizers can make use of it

src/common/Language.ts
```diff
@@ -65,8 +70,7 @@ export function defaultDetokenizer(
   indent = 1
 ): Detokenizer {
   return function (tokenTree: TokenTree): string {
-    // @ts-expect-error
-    const tokens: string[] = [tokenTree].flat(Infinity); // it seems ts doesn't understand flattening the tree
+    const tokens: string[] = finiteFlatten([tokenTree]);
```
MichalMarsalek, line 73:
```suggestion
    const tokens: string[] = finiteFlatten(tokenTree);
```

src/common/Language.ts
```diff
@@ -44,6 +44,11 @@ export type TokenTree = string | TokenTreeArray;
 export type Detokenizer = (tokens: TokenTree) => string;
 export type WhitespaceInsertLogic = (a: string, b: string) => boolean;
 
+export function finiteFlatten(tokenTree: TokenTree): string[] {
+  if (typeof tokenTree === "string") return [tokenTree];
+  return tokenTree.map(finiteFlatten).flat(1);
+}
+
```
MichalMarsalek, line 51: Why is the function called `finiteFlatten`?

This should be implemented in a way where tokens are being directly appended to the final result array without creating the intermediate arrays.

MeWhenI: I named it as opposed to the default flat, which TS seemed concerned would be infinite, renamed to `flattenTree` though. Intermediate array creation fixed.

PR #115, opened 2023-02-13
Add restrictions to frontend

MichalMarsalek:
I expect we will be adding a lot of Nodes and PolygolfOpCode's that are really only intended to be used by the backend plugins for their golfing purposes.
If we restrict what is exposed to the user, we can focus our golfing on what is important instead of supporting 5 different ways to write the same program.
Bans 'POLYGOLF' from user variables so that plugins can introduce new idenrifiers that don't collide with the user ones and are still parseable in unrestricted mode.
The restrictions don't apply to tests.


MeWhenI:
All existing opcodes which aren't on the frontend ops list: `argc` `argv` `argv_get` `gcd` `geq` `rem` `sorted` `trunc_div`. Guessing `geq` was excluded by mistake. `sorted` is a weird one, I can see good arguments for/against including it; personally I'd be in favor of including it, at least for int collections, but I can definitely see the case against it.

MeWhenI (approved):

PR #121, opened 2023-02-15
For argv node

### LINKED ISSUE #90: ForEachArg IR Node
Add an IR node for looping over `argv`. Some languages have unique syntax for this task compared to other cases of looping over arrays (ie. AWK and COBOL) or require things like accessing argc.

MichalMarsalek:
Closes #90 
* Adds `ForArgv` node (and frontend `for_argv`) for looping over the cmd arguments.
* Adds nullary opcode `argc`.
* Adds `forArgvToForEach` plugin
* Adds `forArgvToForRange(true)` plugin that transforms the argv loop to overshooting for range loop.
* Adds `forArgvToForRange(false)` plugin that transforms the argv loop to 1..argc for range loop.
* Adds `assertForArgvTopLevel` plugin for making sure ForArgv is only used at the top level and only once.
* Makes `argv_get` expect an integer literal argument - to prevent users from writing a `for` loop where they should be using the `forArgv` loop while still allowing the op to be used to acces the only input for holes where there's only one (sudoku-v2)
* Integrates the above to Python, Lua, Nim, Swift, Golfscript.
* As a byproduct, implements ForEach loops in the 5 langs and fixes a couple of bugs.


src/languages/polygolf/emit.ts
```diff
@@ -228,6 +228,12 @@ export function emitExpr(
         expr.collection,
         ...emitExpr(expr.body, false, true)
       );
+    case "ForArgv":
+      return emitSexpr(
+        "for_range",
+        expr.variable,
+        expr.argcUpperBound.toString()
```
MeWhenI, line 235:
```suggestion
        "for_argv",
        expr.variable,
        expr.argcUpperBound.toString(),
        ...emitExpr(expr.body, false, true)
```
Fixes polygolf for_argv emit. Is it fine that argcUpperBound isn't an `IR.Expr`?

src/languages/swift/index.ts
```diff
@@ -29,20 +23,9 @@ const swiftLanguage: Language = {
...
-        "argv_get",
-        (x) =>
-          indexCall(
-            id("CommandLine.arguments", true),
-            polygolfOp("add", x[0], int(1n)),
-            "argv_get",
-            true
-          ),
-      ],
-    ]),
```
MeWhenI, line 45: Don't we still need `argv_get` for single argument holes?

src/languages/polygolf/emit.ts
MichalMarsalek: Do you have a usecase for it being `IR.Expr`? I'd say on the frontend it should definitely be an integer so it would have to be a plugin?

PR #123, opened 2023-02-16
Fix #97 & #98

### LINKED ISSUE #97: Python simple block body
**Language**
Python

**Polygolf source**

```
for $i 0 10 {
    $a <- $i;
    println $a;
};
```

**Current output**

```
for i in range(10):
 a=i
 print(a)
```

**Expected output**

```
for i in range(10):a=i;print(a)
```

**Comment**
Code is valid but unnecesarily longer.

### LINKED ISSUE #98: Nim simple block body
**Language**
Nim

**Polygolf source**

```
for $i 0 10 {
    println $i;
    println $i;
};
```

**Current output**

```
for i in..9:
 i.echo
 i.echo
```

**Expected output**

```
for i in..9:i.echo;i.echo
```

**Comment**
Code is valid but unnecesarily longer.

MeWhenI:
Is there a reason I'm not aware of for node to be considered one of its own children? Otherwise, this fixes the issue.

Edit so Github knows which issues this closes:
Fixes #97 
Fixes #98

### DIFF
--- a/src/common/emit.ts
+++ b/src/common/emit.ts
@@ -60,7 +60,7 @@ export function emitStringLiteral(
 }
 
 export function hasChildWithBlock(node: IR.Node): boolean {
-  for (const child of [node, ...getChildren(node)]) {
+  for (const child of getChildren(node)) {
     if ("consequent" in child || "children" in child || "body" in child) {
       return true;
     }
--- a/src/common/emit.ts
+++ b/src/common/emit.ts
@@ -1,5 +1,5 @@
 import { IR } from "IR";
-import { getChildren, PathFragment } from "./fragments";
+import { PathFragment } from "./fragments";
 import { TokenTree } from "./Language";
 
 export function needsParensPrecedence(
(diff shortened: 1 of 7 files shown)


MeWhenI:
I suppose to link to the issue I need to actually write out
Fixes #97 
Fixes #98

MeWhenI:
Converting to draft, I see that clashes with other tests

MeWhenI:
This seems janky. I assert that unless there's a compelling reason why not, any node whos child can be a Block (While, For, If), should have to have a Block as its child, and only Blocks themselves can have non-Block children. So instead of
```
{
 kind="While"
 body={
  kind="If"
 }
}
```
we would have
```
{
 kind="While"
 body={
  kind="Block"
  body=[
    {
     kind="If"
   },
  ]
 }
}
```
. It seems unintuitive that in any case anything other than a Block should be getting passed to `emitBlock`.

MichalMarsalek:
> This seems janky. I assert that unless there's a compelling reason why not, any node whos child can be a Block (While, For, If), should have to have a Block as its child, and only Blocks themselves can have non-Block children. So instead of
> 
> ```
> {
>  kind="While"
>  body={
>   kind="If"
>  }
> }
> ```
> 
> we would have
> 
> ```
> {
>  kind="While"
>  body={
>   kind="Block"
>   body=[
>     {
>      kind="If"
>    },
>   ]
>  }
> }
> ```
> 
> . It seems unintuitive that in any case anything other than a Block should be getting passed to `emitBlock`.

It actually used to have exactly this, and decided to switch to bodies being Exprs.

MichalMarsalek:
It would be nice if the top level separator was `\n` and not `;`.

MeWhenI:
@MichalMarsalek emitBlock refactored to emitMultiExpr which I think more accurately reflects what it does. And yeah gotcha newlines do have a nice look to them.

PR #129, opened 2023-02-17
Fix nim indexing precedence

MichalMarsalek:


MichalMarsalek:
This is completely overwritten in #163 but that can take time to be merged.

jared-hughes (approved):

PR #135, opened 2023-02-18
Implement conditional ops

MeWhenI:
Adds conditional ops to Nim, Lua, and Python.
TODO: Add Tests

### DIFF
--- a/src/languages/lua/emit.ts
+++ b/src/languages/lua/emit.ts
@@ -102,6 +102,9 @@ function needsParens(
   if (needsParensPrecedence(expr, parent, fragment)) {
     return true;
   }
+  if (expr.kind === "ConditionalOp" && expr.isSafe)
+    return parent.kind === "UnaryOp" || parent.kind === "BinaryOp";
+
   if (
     parent.kind === "MethodCall" &&
     expr === parent.object &&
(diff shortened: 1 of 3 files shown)


MichalMarsalek:
I think I would prefer if these were done (where possible) via an emit plugin, instead of hardcoding the array access in the emitter.

MeWhenI:
@MichalMarsalek Each of these langs has 1 version of conditionals that uses array access and 1 that doesnt (safe vs unsafe). Would you think itd be good to do both types of emit in the plugin or just use it to convert to array access when appropriate?

MichalMarsalek:
@MeWhenI 
- New (backend only) OpCode `implicit_int_to_bool` should be added.
- In Python, `mapOps("implicit_int_to_bool", (x) => x[0])` should be added.
- In Nim, `mapOps("implicit_int_to_bool", (x) => polygolfOp("int_to_bool", x[0]))` should be added.
- For unsafe conditional op, if the syntax is sufficiently different from other contructs in the target language, it should be left alone and directly emited in the emitter.
- For safe conditional op, if the shortest form is indexing into and array, a new `unsafeConditionalOpToArrayGet` plugin should be applied.
- This new plugin should be converting `conditional_op $cond $cons $alt;` to `array_get (array $alt $cons) (implicit_int_to_bool $cond);`.

MichalMarsalek:
Alternatively, `unsafeConditionalOpToArrayGet` would just return `array_get (array $alt $cons) (int_to_bool $cond);` and then there would be another plugin, `addImplicitIntToBool` plugin that replaces each explicit conversion with an implicit. I think this is preferred since we might want to apply such plugin anyway - to handle other cases (like multiplication by bool) and also cases where user explicitly uses the `int_to_bool` op.

MichalMarsalek:
I implemented a new `ImplicitConversion` node that could be usedul here.

MichalMarsalek:
I will implement this as part of #201 .

PR #148, opened 2023-02-19
Add and implement bit shift left and right opcodes

### LINKED ISSUE #145: Add bit shift left and bit shift right
Idk why these don't exist.

Steffan153:
Closes #145.


MichalMarsalek:
At least getType tests and Lua tests should be added.

PR #163, opened 2023-02-26
Operators overhaul

### LINKED ISSUE #81: Nim - accidental raw string literal
**Language**
Nim

**Polygolf source**
```
print "Hello\nworld!";
```

**Current output**
```
echo"Hello\nworld!"
```

**Expected output**
```
echo "Hello\nworld!"
```

**Comment**
Not using the space makes the string literal to be interpreted as a raw one.
For the correct behaviour, space should be inserted if and only if the literal contains escape sequences.
On the other hand, if the literal is already raw, the `r` prefix should be removed, i.e.  
`echo r"\n"` should be golfed to `echo"\n"`.

### LINKED ISSUE #83: Add Python tests

### LINKED ISSUE #84: Add Nim tests

### LINKED ISSUE #91: Add Swift tests

### LINKED ISSUE #105: Associative binary ops
Add plugin that regroupschained applications of a associative operator, or integrate this ogic into existing plugins.

### LINKED ISSUE #128: Abstract precedence handling
Generalise the precedence logic instead of all the special casing in the emitters,.

MichalMarsalek:
- This includes #129 #79 - those should be reviewed & merged first.
- (Somewhat) Canonical representation of arithmetic expressions.   
  Associative ops are now represented as `PolygolfOp`s with variadic number of children.
  Polygolf ensures that in the IR, there will never be:
  - Polygolf(neg)
  - Polygolf(sub)
  - Polygolf(add) as a direct child of Polygolf(add) - same with all other associative ops
  - Multiple IntegerLiterals as children of Polygolf(add) or any other commutative ops
  
  This is ensured when using the `polygolfOp` contructor function and the Spine API so avoid creating such nodes manually.
  This closes #105 
- removes precedence and associativity fields from the `BinaryOp` & `UnaryOp` nodes, these properties are determined on emit.
- refactors emit functions by determining a precedence for each node instead of hardcoding all the sepcial cases - closes #128 
 - fixes #81 
 - closes #84, closes #91, closes #83 - many tests were added since I created those issues + I added some tests here


src/IR/exprs.ts
```diff
@@ -123,14 +114,113 @@ export function keyValue(key: Expr, value: Expr): KeyValue {
...
+    return polygolfOp("add", args[0], polygolfOp("neg", args[1]));
+  }
+  if (isBinary(op) && isAssociative(op)) {
+    args = args.flatMap((x) =>
+      x.kind === "PolygolfOp" && x.op === op ? x.args : [x]
+    );
+    if (op === "add") args = simplifyPolynomial(args);
+    else {
+      if (isCommutative(op)) {
+        args.sort(compareTerms);
```
jared-hughes, line 139: It's not clear that the `compareTerms` is correct (it's not anti-symmetric for example). Since this seems to just be putting integer literals at the start, why not something like (untested, and you'll need to also remove the definition of `compareTerms`):
```suggestion
        args = args.filter(x => x.kind === "IntegerLiteral").concat(args.filter(x => x.kind !== "IntegerLiteral"));
```

src/common/getType.ts
```diff
@@ -356,14 +369,23 @@ function getOpCodeType(
...
-      expectType(integerType(), integerType());
-      return getArithmeticType(
-        expr.op,
-        types[0] as IntegerType,
-        types[1] as IntegerType,
-        expr.source
-      );
+    case "max": {
+      const op = expr.op;
+      if (isAssociative(expr.op)) {
```
jared-hughes, line 374: The list of associative ops is fixed. Splitting the block of cases into two seems more natural than having an isAssociative check. I feel I'm missing something.

jared-hughes (approved):
Awesome!

PR #171, opened 2023-03-05
Fix decimal printers

MichalMarsalek:
- Fixes the decimal constant and low decimal list printers.
- Parametrize `useEquivalentTextOp` with two booleans allowing bytes and/or codepoints ops so that it can be used as both Golf plugin and Emit plugin.
- Improve `forRangeToForEach` plugin to work with text literals, list literals and with arrays.
- Add `codepoint_to_int` & `text_byte_to_int` ops, add related plugins & refactor


jared-hughes (commented):
The rest seem good, but perhaps I'm missing something:

> Add `codepoint_to_int` & `text_byte_to_int` ops, add related plugins & refactor

Are these new ops implemented for any lang?

MeWhenI:
Is it intended that `int_to_codepoint` is not a frontend op? Also noting that `int_to_codepoint` will fail in most or all languages for characters in the surrogate pair range `0xd800-0xdfff`, thought maybe this is why its not allowed on frontend.

MichalMarsalek:
> Is it intended that `int_to_codepoint` is not a frontend op? Also noting that `int_to_codepoint` will fail in most or all languages for characters in the surrogate pair range `0xd800-0xdfff`, thought maybe this is why its not allowed on frontend.

`int_to_codepoint` was not added in this PR.

MeWhenI:
right, ill put that up as a separate issue then

PR #172, opened 2023-03-06
Abstract imports adding

### LINKED ISSUE #89: Abstract dependencies
New languages often have
```js
    // get dependencies
    // TODO: abstract this part for other languages
    // TODO: cache, and maybe do recursive merging for performance
```
Need some better way to manage imports.

MichalMarsalek:
Closes #89


src/languages/golfscript/index.ts
```diff
@@ -87,7 +87,13 @@ const golfscriptLanguage: Language = {
...
       ["argv_get", "a="],
     ]),
-    addImports,
+    addImports(
+      [
+        ["a=", "a"],
+        ["a", "a"],
+      ],
+      (x) => (x.length > 0 ? assignment(x[0], id("", true)) : undefined)
+    ),
```
jared-hughes, line 96: What does this do? My guess is it sets up an import for any usage of `a` or `a=`, and the import would be an assignment of `a` to be the empty identifier, so it'll look like `:a`. Oh, is this loading the arguments stack into a variable?

jared-hughes (approved):

src/languages/golfscript/index.ts
MichalMarsalek: > What does this do? My guess is it sets up an import for any usage of `a` or `a=`, and the import would be an assignment of `a` to be the empty identifier, so it'll look like `:a`. Oh, is this loading the arguments stack into a variable?

Thats's correct.

PR #173, opened 2023-03-08
Throw error on negative for range loop in gs

### LINKED ISSUE #141: Golfscript: range doesn't work properly with negatives
`for $b -15 1` compiles to `1,-15>` which obviously doesn't work.

MichalMarsalek:
This closes #141 by throwing an error when the form currently used doesn't work. Not sure what a best general fallback is.

### DIFF
--- a/src/languages/golfscript/emit.ts
+++ b/src/languages/golfscript/emit.ts
@@ -1,133 +1,146 @@
 import { TokenTree } from "../../common/Language";
 import { EmitError, emitStringLiteral } from "../../common/emit";
-import { IR, isIntLiteral } from "../../IR";
+import { Expr, integerType, IR, isIntLiteral, isSubtype, Type } from "../../IR";
+import { getType } from "@/common/getType";
 
 export default function emitProgram(program: IR.Program): TokenTree {
-  return emitStatement(program.body, program);
-}
-
-function emitMultiExpr(baseExpr: IR.Expr, parent: IR.Node): TokenTree {
-  const children = baseExpr.kind === "Block" ? baseExpr.children : [baseExpr];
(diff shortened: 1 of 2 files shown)


src/languages/golfscript/emit.ts
```diff
@@ -1,133 +1,146 @@
...
+              : [emitExpr(stmt.increment), "%"],
+            "{",
+            ":",
+            emitExpr(stmt.variable),
+            ";",
+            emitMultiExpr(stmt.body, stmt),
+            "}",
+            "%",
+          ];
+        throw new EmitError(stmt, "potentially negative");
```
MeWhenI, line 52:
```suggestion
        return [
          isNaturalType(stmt.low)
            ? [
                emitExpr(stmt.high),
                ",",
                isIntLiteral(stmt.low, 0n) ? [] : [emitExpr(stmt.low), ">"],
              ]
            : [
                emitExpr(stmt.high),
                emitExpr(stmt.low),
                "-",
                ",",
                "{",
                emitExpr(stmt.low),
                "+",
                "}",
                "%",
              ],
          isIntLiteral(stmt.increment, 1n)
            ? []
            : [emitExpr(stmt.increment), "%"],
          "{",
          ":",
          emitExpr(stmt.variable),
          ";",
          emitMultiExpr(stmt.body, stmt),
          "}",
          "%",
        ];
```
This handles a negative low and/or high bound. Negative increments function as intended, or at least in the same way as Python. If this is undesired behavior it should probably be impossible to use a negative step regardless of lang. This section of new code:
```
emitExpr(stmt.high),
emitExpr(stmt.low),
"-",
```
should emit a `sub` polyGolfOp acting on high and low, or whatever the best way is to enable the subtraction to be simplified, I'm not sure how to go about that though.

PR #174, opened 2023-03-09
Add ForDifferenceRange & use it for gs neg loops

MichalMarsalek:
- Adds new `ForDifferenceRange` node for loops that specify start and a different between start and end
- Renames range's low/high to start/end since step can be negative
- Fixes type inference for ranges with negative steps
- Adds tests


MeWhenI:
An issue seems to arise from `forRangeToForDifferenceRange` being a final emit plugin and generating a `sub` polygolfOp. It appears that `IR.exprs` simplifies `sub(x,y)` to `add(x,neg(y)` and then it is expected for a golf plugin to come in and clean that up. This doesn't work for the diff range plugin, I assume because golf plugins aren't run after final emit plugins.
 
```polygolf
$a:-10..10 <- -4;
for $i $a 1 {
  println $i;
};
```
compiles to gs as 
```golfscript
-4:a;a-1*1+,{a+:i;i puts}%
```
rather than the preferable
```golfscript
-4:a;1 a-,{a+:i;i puts}%
```
Otherwise looks good :thumbsup:

MichalMarsalek:
That issue is actually unrelated to the PR,
```
$a:-10..10 <- -4;
println (1 - $a);
```
compiles to
```
-4:a;a-1*1+puts
```
The emit plugin `mapToUnaryAndBinaryOps` should be improved to better handle these.

PR #181, opened 2023-04-15
Emit negative polynomial terms last

MichalMarsalek:
- Patches `mapToUnaryAndBinaryOps` so that negative terms in a polynomial are always placed last.
- Enforces the invariant
> IntegerLiteral (if present) appears at the first position in the args list

for `polygolfOp` of type `add` as well.

### DIFF
--- a/src/IR/exprs.ts
+++ b/src/IR/exprs.ts
@@ -38,7 +38,7 @@ export interface KeyValue extends BaseExpr {
  * - PolygolfOp(neg)
  * - PolygolfOp(sub)
  * - PolygolfOp as a direct child of a PolygolfOp with the same associative OpCode
- * - IntegerLiteral as a nonfirst child of an associative PolygolfOp
+ * - IntegerLiteral as a nonfirst child of a commutative PolygolfOp
  * 
  * This is ensured when using the polygolfOp contructor function and the Spine API so avoid creating such nodes manually.
  */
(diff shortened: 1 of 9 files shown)


src/IR/exprs.ts
```diff
@@ -173,7 +173,9 @@ export function polygolfOp(op: OpCode, ...args: Expr[]): Expr {
         } else newArgs.push(arg);
       }
       args = newArgs;
-      if (
+      if (op === "mul" && args.length === 2 && isIntLiteral(args[0], 1n)) {
```
MeWhenI, line 176:
```suggestion
      if (op === "mul" && args.length > 1 && isIntLiteral(args[0], 1n)) {
```
Removes redundant multiplication by 1 for arbitrarily sized multiplication chains, as currently
```
$a:-5..5 <- 0;
$b:-5..5 <- 0;
$c <- (* (- $a) (- $b));
$d <- (* 1 $a $b);
```
compiles to 
```
$a:-5..5 <- 0;
$b:-5..5 <- 0;
$c <- (* 1 $a $b);
$d <- (* 1 $a $b);
```

MichalMarsalek:
Also improved handling of the multiplicative -1, so that
```
$a:Int <- 0;
print ((- $a) * ($a - 3));
```

is

```
a=0
print(a*(3-a))
```
in Python, it used to be
```
a=0
print(-(a*(a-3)))
```

PR #182, opened 2023-04-15
One step loops

### LINKED ISSUE #176: Add plugin converting to 1-step loops
Transform

```
for $i 0 10 2 {println $i;};
```

into 

```
for $j 0 5 {$i <- (2*$j); println $i;};
```

(and #119 can then convert it to `for $i 0 5 {println (2*$i);};`)

MichalMarsalek:
Closes #176


MeWhenI:
At least one of either `forRangeToForRangeInclusive` or `forRangeToForRangeOneStep` should be an emit plugin rather than a golf plugin in Nim since Nim can't emit an Exclusive For Loop with `step != 1`.

MichalMarsalek:
> At least one of either `forRangeToForRangeInclusive` or `forRangeToForRangeOneStep` should be an emit plugin rather than a golf plugin in Nim since Nim can't emit an Exclusive For Loop with `step != 1`.

We couldn't have just applied `forRangeToForRangeInclusive` as emit plugin since that would mean that we *never* get the exclusive variant. Same with `forRangeToForRangeOneStep`. I solved it by adding a parameter to `forRangeToForRangeInclusive` that determines whether it should also be applied to the case where step==1.

Also, note that if some path of the search yields an IR that cannot be emitted, that isn't really an issue, as long as the search can reach a different IR that can be.

MeWhenI:
Yeah for sure, I understand not every path needs to be emittable. Looks all good 👍

PR #183, opened 2023-04-16
Make loop variables readonly

### LINKED ISSUE #125: Readonly loop variables
Make loop variables readonly (perhaps by extending the type system?).

MichalMarsalek:
Closes #125 .


MeWhenI:
I believe there may be cases where it is desirable to make exceptions to this for some langs down the road, but looks good for the time being.

PR #195, opened 2023-04-25
Remove op fields from every node but PolygolfOp

### LINKED ISSUE #188: Get rid of `op` fields on presentational nodes
The `op` fields on `BinaryOp`, `UnaryOp`, `FunctionCall` and other emit phase nodes should be removed. They are not needed for emit and semantic transformations should be performed when these are still in the form of `PolygolfOp`s.
Note that `FunctionCall` is a front end node as well but in this case, the `op` field is not used either.

MichalMarsalek:
Closes #188


MeWhenI (approved):
Looks good 👍

PR #196, opened 2023-04-25
remove evalStaticExpr

### LINKED ISSUE #189: Get rid of `evalStaticExpr`
Polygolf now ensures that the IR for a `PolygolfOp` is somewhat normalized and so this plugin shouldn't be needed.

MichalMarsalek:
Closes #189

### DIFF
--- a/src/languages/golfscript/index.ts
+++ b/src/languages/golfscript/index.ts
@@ -17,7 +17,6 @@ import {
   flipBinaryOps,
 } from "../../plugins/ops";
 import { renameIdents } from "../../plugins/idents";
-import { evalStaticExpr } from "../../plugins/static";
 import { golfLastPrint } from "../../plugins/print";
 import {
   forArgvToForEach,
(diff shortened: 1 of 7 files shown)


MeWhenI:
Assuming we're fine for now with some ops that previously would've been simplified no longer being, like `mod` and `div`, looks good.

MichalMarsalek:
> Assuming we're fine for now with some ops that previously would've been simplified no longer being, like `mod` and `div`, looks good.

Oh crap, I forgot that the change that makes `polygolfOp` simplify those ops is not in main yet, it is in #182 .

MichalMarsalek:
> > Assuming we're fine for now with some ops that previously would've been simplified no longer being, like `mod` and `div`, looks good.
> 
> Oh crap, I forgot that the change that makes `polygolfOp` simplify those ops is not in main yet, it is in #182 .

I pushed the relevant part to main and so I'm merging this now.

PR #197, opened 2023-04-27
Typed print ops

### LINKED ISSUE #114: Print ops
* Typecheck `print` and `println` to expect single Text argument.
* Add `print_int`, `println_int` opcodes.
* Add `println_list_joined_using` opcode expecting a list and text.
* Add `println_many_joined_using` opcode expecting a arbitrary many arguments and then one text argument
* Add related plugins.
* Remove typecheck for print in Python.
* Add `print` and `println` to UnaryOpCode array & type.

MichalMarsalek:
Closes #114


src/common/getType.ts
```diff
@@ -275,8 +275,11 @@ function getOpCodeType(
   program: Program
 ): Type {
   const types = getArgs(expr).map((x) => getType(x, program));
-  function expectVariadicType(expected: Type) {
-    if (types.length < 2 || types.some((x, i) => !isSubtype(x, expected))) {
+  function expectVariadicType(expected: Type, minArity = 2) {
```
MeWhenI, line 278: why default minArity to 2?

MeWhenI:
Looks good, do you want to the joining print operators implemented before pulling?

src/common/getType.ts
MichalMarsalek: > why default minArity to 2?

That was the previous behaviour. When I add new parameters to a function I tend to add a default value that keeps the old semantics.

MichalMarsalek:
> Looks good, do you want the joining print operators implemented before pulling?

I don't.

PR #201, opened 2023-04-29
Conditional ops, short circuiting, relation chains

### LINKED ISSUE #100: Use shortcircuiting
Add plugin which golfs

```python
if not x:print(y)
```

to

```python
x or print(x)
```

MichalMarsalek:
Closes #100
Golfs
- [x] `if x:print(y)` to `x and print(x)` in Python
- [x] `if x==y:print(z)` to `x==y==print(x)` in Python
- [x] `x if c else y` to `[y,x][c]` in Python, to `[y,x][int c]` in Nim, to `c and x or y` in Lua
- [x] `y<z and x<y` to `x<y<z` in Python

Bullets 1 and 2 are currently implemented only for print as I wasn't sure how to generalise that.


MeWhenI:
Precedence appears to be incorrect in Nim, as
```polygolf
$a:Int <- 0;
$b:Int <- 0;
println_int (conditional ($a > 0) (conditional ($b > 0) 1 2) 3);
```
compiles to code that produces an error:
```Nim
var a,b=0
if a>0:[2,1][int b>0]else:3.echo
```
It should compile to something like
```Nim
var a,b=0
(if a>0:[2,1][int b>0]else:3).echo
```

MeWhenI:
Compiling
```polygolf
$a:Int <- 1;
$b:Int <- 1;
$c:Int <- 1;
print_int (conditional (conditional ($a > 0) ($b > 0) ($c > 0)) 8 7);
```
to Swift produces
```Swift
var a=1,b=1,c=1
print(a>0 ?b>0:c>0 ?8:7)
```
where it should produce
```Swift
var a=1,b=1,c=1
print((a>0 ?b>0:c>0) ?8:7)
```

MeWhenI:
This code should print "pass":
```polygolf
$a <- (conditional true false true);
if (not $a) {println "pass";};
```
It does not do so when compiled to Lua:
```lua
a=true and false or true
if not a then print("pass")end
```

MeWhenI:
Compiling
```polygolf
println_int (conditional (conditional true false true) 6 (conditional true 7 8));
println_int (conditional true (conditional false 8 7) 6);
println_int (conditional true (conditional true 7 8) 6);
```
to Python produces
```py
p=print
p([[8,7][1],6][[1,0][1]])
p([6,[7,8][0]][1])
p([6,[8,7][1]][1])
```
which is the desired (shortest) output. However if the last println is changed to a print, ie.
```polygolf
println_int (conditional (conditional true false true) 6 (conditional true 7 8));
println_int (conditional true (conditional false 8 7) 6);
print_int (conditional true (conditional true 7 8) 6);
```
it instead compiles to
```py
p=print
p(6 if[1,0][1]else[8,7][1])
p([6,[7,8][0]][1])
p([8,7][1]if 1 else 6)
```
Working on finding a shorter minimum reproducible example but its very finicky.

MeWhenI:
Compiling 
```polygolf
print_int (conditional true 7 8);
println_int 1;
```
to Nim produces
```Nim
stdout.write[8,7][true.int]
1.echo
```
Whitespace is needed between `write` and the square bracket

MeWhenI:
I've been bug testing this PR by compiling
```polygolf
print "Expect all 7's: ";
print_int (conditional true 7 8);
print_int (conditional false 8 7);
print_int (conditional (conditional true false true) 8 7);
print_int (conditional (conditional true false true) 6 (conditional true 7 8));
print_int (conditional true (conditional false 8 7) 6);
print_int (conditional true (conditional true 7 8) 6);
print_int (conditional (conditional true false true) (conditional false 6 4) (conditional true 7 8));
```
to the different langs and checking the result. Doing so reveals buggy nested conditionals in Python and Nim. However these shouldnt be outputting as conditionals in the first place, they should converted to array indexing. Compiling any individual line of this code on its own seems to succesfully convert it to array indexing in those languages.

MichalMarsalek:
Damn what a buggy PR! Thank you for the thorough review/testing.

MichalMarsalek:
This PR is really outdated. I will be redoing it in a couple of smaller PRs.

PR #217, opened 2023-05-07
Make all nodes parseable

### LINKED ISSUE #194: Make more (all?) nodes parseable in unrestricted mode

MichalMarsalek:
Closes #194 
I had to change the grammar a bit, so that every node can be a child of a block.
I separated PropertyCall node from a MethodCall node.
The order in which the parser expects the children matches the order in the constructor functions (like `functionCall`) expects them.


src/plugins/ops.ts
```diff
@@ -213,3 +215,12 @@ export const removeImplicitConversions: Plugin = {
     }
   },
 };
+
+export const methodsAsFunctions: Plugin = {
+  name: "removeImplicitConversions",
```
MeWhenI, line 220:
```suggestion
  name: "methodsAsFunctions",
```

MeWhenI (approved):

PR #218, opened 2023-05-20
Text replace op

MichalMarsalek:
- Implements the `text_replace` op in Lua, Nim, Python, Swift. In Lua, this is done by escaping the arguments and using the `gsub` function, I don't know if there's a better way.
- Implements a new backend-only `text_multireplace` op in Nim and Python.
- Adds a plugin converting a nested `text_replace`s to `text_multireplace`.

### DIFF
--- a/src/IR/opcodes.ts
+++ b/src/IR/opcodes.ts
@@ -217,7 +217,7 @@ export const OpCodes = [
   "argv",
   "argc",
   "text_replace",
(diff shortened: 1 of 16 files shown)


MeWhenI:
Small bugfix in #219, I thought it would show up in this thread rather than in its own PR. But yeah looks good besides that fix.

PR #221, opened 2023-05-25
Inlining variables + stdin

### LINKED ISSUE #119: Variable inlining
Inline variables that are only used once, so that

```
$a <- (argv_get 0);
$b <- (text_length $a);
$c <- ($b * 3);
print (int_to_text $c);
```

compiles to Python as

```python
import sys
print(len(sys.argv[1])*3)
```

This makes writing complex expressions easier in Polygolf by allowing naming the intermediate steps.

MichalMarsalek:
- Adds ops for reading from stdin. This was mainly motivated by the fact that inputs to some langs on code.golf are provided trought stdin. However, I added it for langs to which that doesn't apply on code.golf because it's almost no work.
A language can set `readsFromStdinOnCodeDotGolf: true`, by default it is assumed that input is from cmd args. When compiling a multivariant Polygolf code to a language, only variants that use the matching input method to code.golf are considered, unless there is no such variant, in which case all variants are considered. This makes it so that we primarily target code.golf by default, but other targets are also possible.
- Adds node classifications and testing functions for:
  - Writing access of a variable
  - Reading access of a variable
  - Side effect
  - Reading input
- Uses that to close #119: Variable is inlined if it is only written to once and if the right side of the assignment doesn't cause a side effect.


src/languages/nim/index.ts
```diff
@@ -96,7 +96,7 @@ const nimLanguage: Language = {
       useMultireplace(),
       inlineVariables,
       forArgvToForEach,
-      forArgvToForRange(),
+      forArgvToForRange(true),
```
btnlq, line 100: `forArgvToForRange(true, true)` ?

MichalMarsalek: With the `forRangeToForRangeInclusive` plugin included, it should not matter.

PR #224, opened 2023-05-28
Join op improvements

### LINKED ISSUE #220: Plugin to convert text replace to split + join
Add a plugin to convert
```polygolf
 text_replace $a $b $c
```
to
```polygolf
join_using (text_split $a $b) $c
```

This can sometimes be shorter in Swift, and is useful for golfscript which doesn't have a built-in text replace function

MichalMarsalek:
- Merges `join` & `join_using` into a single op
- Implements `join` in the rest of the langs
- Closes #220


MeWhenI:
I PR'ed a few extra tests in #237 for Swift, Nim, & Lua. Otherwise looks all good.

PR #230, opened 2023-08-04
Improved search / language definition

### LINKED ISSUE #150: Add new kinds of plugins to Language definition
1. Add `preprocessPlugins` (or `presearchPlugins`) array to `Language`.
Plugins in this array would run once in the specified order before the search starts (before `golfPlugin`s). They would be intended for
a) Plugins that should only be ran once.
b) Plugins that can run multiple times but running them first will help guide the search and make it more efficient.

2. Add `preemitPlugins` (or `postsearchPlugins`) array to `Language`.
Plugins in this array would run once in the specified order after the search but before the emit phase (before running `emitPlugins` as part of emit).
Plugins that never make the code longer should be always applied and as such should currently be placed in `finalEmitPlugins` or into `emitPlugins` if there's a suspicion that additionally running them earlier might help the search phase.
However this doesn't make sense since in either case they *might not be* emit plugins in their nature - meaning that not applying them doesn't make the output invalid. This is relevant for testing where I might expect polygolf `$apple <- 3;` with the `python nogolf` to match `apple=3` and not the current `a=3`. This is a silly example, but adding more and more plugins to `emitPlugins` or `finalEmitPlugins` arrays that really just perform some sort of golf can make it harder to test the emitters since even with `nogolf` option, the nodes would just change too much.

3. I'm not really feeling the `emitPlugins` vs `finalEmitPlugins` distinction. I think that plugin should either be used as golfing one or as an emit one. In cases where it makes sense, one can always explicitly list the same plugin in both `golfing` array and `emit` (= current `finalEmit`) array.

MichalMarsalek:
Closes #150 

In addition to the original proposition, I decided to implement the option of multiple search phases.
I think that mostly, using a single one would suffice, but I have some ideas of where second one (that operated on a emitter supported set of nodes) might be useful.

- Adds support for compilation warnings - each plugin and an emitter are exposed to a `addWarning` function. Warning can local to the search path or global (stays even if the search path is not the shortest).
- Collects all compilation parameters to a single `CompilationOptions` object.
- Each plugin and emitter are exposed to the compilation options. This can be useful for slightly different emit for example of text literals in char vs byte target, or to disable certain plugin in some situation as a performance improvement.
- Tracks the history of plugins applied along each search path, along with information of whether they actually performed a replacement or not.
- Makes all languages output the default when no lang is specified.
- Adds new `all` switch to CLI to get outputs for all variants rather than just the shortest.
- Adds new `debug` switch to CLI to get some debug info (currently only the history of plugins).
- Moves the testing logic to a separate file instead of generating it for each md test.
- Make all the fields of `CompilationOptions` configurable in the markdown tests using keywords after the name of the lang. All arguments that are not from the set of keywords are interpreted as plugins (so there can be multiple now).


package-lock.json
```diff
@@ -9,6 +9,7 @@
       "version": "0.1.0",
       "license": "MIT",
       "dependencies": {
+        "@datastructures-js/priority-queue": "^6.3.0",
```
jared-hughes, line 12: Nice find! Whenever I need a pq in js, I always end up copying from some medium post lol.

src/cli.ts
```diff
@@ -37,44 +34,68 @@ const options = yargs()
...
-        console.log(result);
-      }
-    } catch (e) {
-      handleError(e);
+    } else {
+      console.log(result.result);
+    }
+    if (result.warnings.length > 0) {
+      console.log("Warnings:");
+      console.log(result.warnings.map((x) => x.message).join("\n"));
```
jared-hughes, line 87: Should warnings be printed only if `options.debug === true`?
Should these go to stderr? (same with the other debugs down below).

src/common/Language.ts
```diff
@@ -64,7 +89,11 @@ export interface IdentifierGenerator {
   general: (i: number) => string;
 }
 
-export type Emitter = (program: IR.Program) => TokenTree;
+export type Emitter = (
+  program: IR.Program,
+  addWarning: AddWarning,
+  compilationOptions: CompilationOptions
```
jared-hughes, line 95: Could be reasonable to combine `addWarning` and `compilationOptions` into a single "Context" object if they're often passed around together.

src/common/compile.ts
```diff
@@ -0,0 +1,493 @@
...
+    } else {
+      len++;
+    }
+  }
+
+  return len;
+};
+
+export const byteLength = (x: string | null) =>
+  x === null ? Infinity : Buffer.byteLength(x, "utf-8");
```
jared-hughes, line 61: This is a large file, and `charLength` and `byteLength` are really just utility functions -- should be split off to another file.

Perhaps also split off `shorterBy` too (anything else?) to theme the helper file around objective functions.

src/common/compile.ts
```diff
@@ -0,0 +1,493 @@
...
+      addWarning,
+      options,
+      ...finishingPlugins.map((x) => x.visit)
+    );
+    return [
+      emit(language, resProg, addWarning, options),
+      finishingPlugins.map((x, i) => [counts[i], x.name]),
+    ];
+  }
+  let shortestSoFar: SearchState;
```
jared-hughes, line 316: If you type this as `let shortestSoFar: SearchState | undefined;`, then you don't need any of the `shortestSoFar!` assertions.

jared-hughes (commented):
Nice! I like the direction. Only skimmed it, but it looks good.

src/cli.ts
MichalMarsalek: My idea was that the debug option is for debugging Polygolf, not really for debugging the Polygolf programs. But I guess it also depends on how the warnings will actually be used, there are currently none being generated.

src/common/Language.ts
MichalMarsalek: True, this also came in mind.

src/common/compile.ts
MichalMarsalek: I tried many different version around this and I was encountering some very weird behaviours that I couldn't understand and considered them possible bugs. This ugly thing was the only thing that worked for me. :(

MichalMarsalek: Never mind, your suggestion apparently works.
I guess, I was only trying `let shortestSoFar: SearchState | undefined = undefined;` which doesn't work as typescript narrows the variable type to just `undefined` in the section after the `while`. Like it doesn't understand that the `enqueue` closes over the `shortestSoFar` variable or something....

src/cli.ts
MichalMarsalek: Maybe, instead of a boolean flag for debug, we could have a verbosity scale.

PR #254, opened 2023-10-30
Nim function calls

MichalMarsalek:
Nim has at least 5 different syntaxes for calling a function. Up until now there was some heuristics that always chose one of those.
This PR uses a second search phase (implemented in #230 ) to chose the shortest form of each function call as the very last step of the node transformations.
The default "nogolf" form is now the `f(x)` syntax, the `f x` & `x.f` are represented as binary ops with names `" "` & `"."` and are being suggested by the `useUFCS` plugin.


MeWhenI:
I'm not certain enough on Nim function precedence to provide the most helpful double-check here but the logic makes sense and I can't seem to come up with any edge cases to break it so looks good from me

PR #255, opened 2023-10-31
Conditional expressions

MichalMarsalek:
- Implements conditional expressions in Lua, Nim, Python, fixes precedence in Swift.
- Adds `bakeType` flag to plugins. (Should this not be a default behaviour?)


MeWhenI:
JS, and by extension Text lang, improperly handles a conditional in the first argument of another conditional
```
print_int (conditional (conditional (1 > 0) (1 < 0) (1 > 0)) 1 2);
```
Compiling to JS produces 
```js
write(1>0?1<0:1>0?1:2)
```
which outputs `false`.
It should produce 
```js
write((1>0?1<0:1>0)?1:2)
```
which outputs `2`.

MeWhenI:
Compiling this test program to Nim
```
print_int (conditional (1 < 0) 1 9);
print_int (conditional (1 > 0) 9 1);
print_int (conditional (1 < 0) (conditional (1 < 0) 1 2) 9);
print_int (conditional (1 < 0) (conditional (1 > 0) 1 2) 9);
print_int (conditional (1 > 0) (conditional (1 < 0) 1 9) 2);
print_int (conditional (1 > 0) (conditional (1 > 0) 9 1) 2);
print_int (conditional (1 < 0) 1 (conditional (1 < 0) 2 9));
print_int (conditional (1 < 0) 1 (conditional (1 > 0) 9 2));
print_int (conditional (1 > 0) 9 (conditional (1 < 0) 1 2));
print_int (conditional (1 > 0) 9 (conditional (1 > 0) 1 2));
print_int (conditional (conditional (1 > 0) (1 < 0) (1 > 0)) 1 9);

% grab from argv to prevent the program being golfed to
% a print statement
$zz <- (argv_get 0);
```
produces
```nim
import os
stdout.write[9,1][int 1<0]
stdout.write if 1>0:9 else:1
stdout.write if 1<0:if 1<0:1 else:2 else:9
stdout.write if 1<0:if 1>0:1 else:2 else:9
stdout.write if 1>0:if 1<0:1 else:9 else:2
stdout.write if 1>0:if 1>0:9 else:1 else:2
stdout.write if 1<0:1 else:if 1<0:2 else:9
stdout.write[if 1>0:9 else:2,1][int 1<0]
stdout.write[if 1<0:1 else:2,9][int 1>0]
stdout.write[if 1>0:1 else:2,9][int 1>0]
stdout.write if[1>0,1<0][int 1>0]:1 else:9
var z=1.paramStr
```
This code has quite a few syntax errors and seems to use the longer `if` syntax in many places where it could use shorter array syntax.

MeWhenI:
This PR appears to have no issues in golfscript, Lua, Python, or Swift

MichalMarsalek:
I fixed the emit problems in Nim, but the fact that only some conditionals are rewritten to arrays might be a problem in the search itself. Removing all plugins except for `safeConditionalOpToCollectionGet("array")` from the search phase leads to more conditionals being rewritten, but not still not all.
The question is, whether this should be use in a search at all, or if it is always shorter.

MeWhenI:
The fixes look good, as for whether array is always shorter in Nim, it sure looks to be but idk enough Nim to be sure. I realized I didn't fully check out `unsafe_conditional`, and Python has a small bug there, same as what JS had with a conditional as the first arg of a conditional. Compiling
```
println_int (unsafe_conditional (unsafe_conditional (1 > 0) (1 < 0) (1 > 0)) 1 2);
```
to Python produces
```py
print(1 if 1<0 if 1>0 else 1>0 else 2)
```
It should produce
```py
print(1 if (1<0 if 1>0 else 1>0) else 2)
```

MichalMarsalek:
> The fixes look good, as for whether array is always shorter in Nim, it sure looks to be but idk enough Nim to be sure. I realized I didn't fully check out `unsafe_conditional`, and Python has a small bug there, same as what JS had with a conditional as the first arg of a conditional. Compiling
> 
> ```
> println_int (unsafe_conditional (unsafe_conditional (1 > 0) (1 < 0) (1 > 0)) 1 2);
> ```
> 
> to Python produces
> 
> ```python
> print(1 if 1<0 if 1>0 else 1>0 else 2)
> ```
> 
> It should produce
> 
> ```python
> print(1 if (1<0 if 1>0 else 1>0) else 2)
> ```

Thank you. Funnily enough, I fixed the precedence of the conditional ternary in Swift as a part of the PR, but at the same time introduced the same bug in other lagnuages....

MichalMarsalek:
Ah, but I didn't realise Python needs the parens there since it's unambiguous without them.

Issue #137, opened 2023-02-19
Add plugin to automatically use the walrus operator (or inline assignment)


Steffan153:
```python
a=2;print(a)
```
should automatically become
```python
print(a:=2)
```
In some other langs it would just become `print(a=2)`.

PR #258, opened 2023-11-02
Javascript

### LINKED ISSUE #180: Introduce the concept of "target types"
This is especially needed for int vs bigint distinction but can be useful for other cases like char vs length-1 string.

MichalMarsalek:
- Closes #180 .
- Implements Javascript.

### DIFF
--- a/src/IR/IR.ts
+++ b/src/IR/IR.ts
@@ -58,7 +58,7 @@ export interface BaseNode {
   /** type: an uninferrable type, either annotated from the frontend or
    * inserted for language-specific op nodes */
   readonly type?: Type;
-  readonly targetType?: string;
+  readonly targetType?: "bigint" | "int" | "string" | "char" | (string & {});
 }
 
 export interface SourcePointer {
--- a/src/IR/IR.ts
+++ b/src/IR/IR.ts
@@ -58,6 +58,7 @@ export interface BaseNode {
   /** type: an uninferrable type, either annotated from the frontend or
    * inserted for language-specific op nodes */
   readonly type?: Type;
+  readonly targetType?: string;
 }
 
 export interface SourcePointer {
--- a/src/IR/opcodes.ts
+++ b/src/IR/opcodes.ts
@@ -3,8 +3,8 @@ export const FrontendOpCodes = [
   "sub",
   "mul",
   "div",
-  "pow",
   "mod",
+  "pow",
   "bit_and",
   "bit_or",
   "bit_xor",
@@ -57,7 +57,6 @@ export const FrontendOpCodes = [
   "text_byte_length", // Returns the text length in bytes.
   "text_codepoint_length", // Returns the text length in codepoints.
   "text_split_whitespace",
-  "join",
   "text_byte_reversed", // Returns a text containing the reversed order of bytes.
   "text_codepoint_reversed", // Returns a text containing the reversed order of codepoints.
   "text_byte_to_int",
--- a/src/languages/javascript/javascript.test.md
+++ b/src/languages/javascript/javascript.test.md
@@ -1,121 +1,105 @@
 # Javascript
 
-## Printing
+## Ops emit
 
 ```polygolf
-println_int 1;
-print_int 2;
-print "3";
-print_int 4;
+$t:(Ascii 3) <- "";
+$n:0..1 <- 0;
+$m <- $n;
+$b <- true;
+$L:(List Int) <- (list 3 2);
+
+$m + $n;
+$m - $n;
+$m * $n;
+$m div $n;
+$m mod $n;
+$m ^ $n;
+$m & $n;
+$m | $n;
+$m ~ $n;
+min $m $n;
+max $m $n;
+$m < $n;
+$m <= $n;
+$m == $n;
+$m != $n;
+$m > $n;
+$m >= $n;
+
+or $b $b;
+and $b $b;
+
+list_length $L;
+list_contains $L 3;
+list_get $L 0;
+list_push $L 58;
+list_find $L 3;
+
+concat $t"bc";
+repeat "x" 10;
+text_contains $t "sub";
+text_replace $t " " "-";
+text_split $t " ";
+join (list $t "www") ",";
+join (list $t "www") " ";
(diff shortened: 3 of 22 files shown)


src/IR/IR.ts
```diff
@@ -58,6 +58,7 @@ export interface BaseNode {
   /** type: an uninferrable type, either annotated from the frontend or
    * inserted for language-specific op nodes */
   readonly type?: Type;
+  readonly targetType?: "bigint" | "int" | "string" | "char" | (string & {});
```
jared-hughes, line 61: Nice hack with the `& {}`. I've seen it before, so I remember, but for your sanity might want to write what it does.

src/IR/opcodes.ts
```diff
@@ -3,8 +3,8 @@ export const FrontendOpCodes = [
   "sub",
   "mul",
   "div",
-  "pow",
   "mod",
+  "pow",
```
jared-hughes, line 7: Does the order matter?

src/IR/IR.ts
jared-hughes: For languages like C, would it make sense to have separate `"int32"` and `"int64"` target types? Since int64 requires `long` which is one byte longer.

src/languages/javascript/javascript.test.md
```diff
@@ -0,0 +1,231 @@
...
+text_contains "A" "subA";
+text_contains "B" "subB";
+text_contains "C" "subC";
+```
+
+```js
+i="includes"
+"A"[i]`subA`
+"B"[i]`subB`
+"C"[i]`subC`
```
jared-hughes, line 167: Are there any existing plugins to allow this?

```js
"A"[i="includes"]`subA`
"B"[i]`subB`
"C"[i]`subC`
```

src/languages/javascript/javascript.test.md
```diff
@@ -0,0 +1,231 @@
...
+$y: 0..100 <- 2;
+println_int ((3 + $x) * $y);
+$t <- (int_to_text $x);
+```
+
+```js nogolf
+x=5n
+y=2
+print((3n+x)*BigInt(y))
+t=String(x)
```
jared-hughes, line 218: I think `""+x` is the same as `String(x)` in all cases.

src/languages/javascript/javascript.test.md
```diff
@@ -0,0 +1,231 @@
...
+## Fixed length for loop
+
+```polygolf
+for $i 25 {
+  println_int $i;
+};
+```
+
+```js
+for(i in{}+1e9)print(i)
```
jared-hughes, line 230: beautiful

src/plugins/ops.ts
```diff
@@ -236,6 +243,19 @@ export function addMutatingInfix(
   };
 }
 
+export const addPostfixIncAndDec: Plugin = {
+  name: "addPostfixIncAndDec",
+  visit(node) {
+    if (
+      node.kind === "MutatingInfix" &&
+      ["+", "-"].includes(node.name) &&
+      isIntLiteral(1n)(node.right)
```
jared-hughes, line 252: Note `x+=-1` and `x-=-1` work like `--x` and `++x` too.

src/plugins/ops.ts
```diff
@@ -236,6 +243,19 @@ export function addMutatingInfix(
   };
 }
 
+export const addPostfixIncAndDec: Plugin = {
+  name: "addPostfixIncAndDec",
+  visit(node) {
+    if (
+      node.kind === "MutatingInfix" &&
+      ["+", "-"].includes(node.name) &&
```
jared-hughes, line 251: Note the return value of `(++x)` is the same as the return value of `(x+=1)` and `(x++)+1`. So prefix might be more appropriate -- I'm not sure where this gets used.

src/plugins/types.ts
```diff
@@ -22,3 +37,64 @@ export const assertInt64: Plugin = {
...
+  primitiveIntType0: "int64" | "int53" | IntegerType,
+  allowed: Partial<Record<OpCode | Assignment["kind"], "bigint" | "int">>,
+): Plugin {
+  const primitiveIntType = type(primitiveIntType0) as IntegerType;
+  return {
+    name: "floodBigints",
+    visit(node, spine) {
+      let nodeType: Type;
+      try {
+        nodeType = getType(isAssignment(node) ? node.variable : node, spine);
```
jared-hughes, line 51: When can this throw?

src/languages/javascript/javascript.test.md
MichalMarsalek: Not yet, but we are [tracking it](https://github.com/polygolf-lang/polygolf/issues/137).

src/IR/opcodes.ts
MichalMarsalek: I don't remember why I changed it.

src/IR/IR.ts
MichalMarsalek: > For languages like C, would it make sense to have separate `"int32"` and `"int64"` target types? Since int64 requires `long` which is one byte longer.

Sure, but that can be added when it is first used.

src/languages/javascript/javascript.test.md
MichalMarsalek: Taken from https://code.golf/wiki/langs/javascript .

src/plugins/ops.ts
MichalMarsalek: `mapToPrefixAndInfix` / `addMutatingInfix` never creates those.

src/plugins/ops.ts
MichalMarsalek: I was thinking we will need dedicated node for inc/dec to handle more golfing anyway. Currently the return is always discarded.

src/plugins/types.ts
MichalMarsalek: Apparently never in tests. I copied it from `assertInt64` but this plugin is run before nodes whose type cannot be calculated are introduced whereas `assertInt64` is ran at the very end.

PR #263, opened 2023-11-11
Temp interpreter

MichalMarsalek:
- Adds a temporary interpreter by running eval on an output of a (modified version of) javascript target.
  - The modified version adds variable declarations and performs instrumentation of loops, so that an infinite loop doesn't leave Polygolf hanging.
- Adds [Text language](https://esolangs.org/wiki/Text) (so user can request output of a Polygolf program that doesn't read input from CLI or playground).
- Adds `hardcode` plugin that transforms any program that doesn't read input to `print "hardcoded output";`.

### DIFF
--- a/src/common/compile.ts
+++ b/src/common/compile.ts
@@ -26,6 +26,7 @@ export interface CompilationOptions {
   skipTypecheck: boolean;
   restrictFrontend: boolean;
   codepointRange: [number, number];
+  skipPlugins: string[];
 }
 
 export function compilationOptions(
@@ -38,6 +39,7 @@ export function compilationOptions(
     skipTypecheck: partial.skipTypecheck ?? false,
     restrictFrontend: partial.restrictFrontend ?? true,
     codepointRange: partial.codepointRange ?? [1, Infinity],
+    skipPlugins: [],
   };
 }
 
@@ -297,7 +299,10 @@ export function compileVariantNoPacking(
   options: CompilationOptions,
   language: Language,
 ): CompilationResult {
-  const phases = language.phases;
+  const phases = language.phases.map((x) => ({
+    mode: x.mode,
+    plugins: x.plugins.filter((x) => !options.skipPlugins.includes(x.name)),
+  }));
   if (
     phases.length < 1 ||
     options.level === "nogolf" ||
@@ -364,7 +369,7 @@ export function compileVariantNoPacking(
     history: [number, string][],
     warnings: Error[],
   ) {
-    if (startPhase >= language.phases.length) return;
+    if (startPhase >= phases.length) return;
     if (latestPhaseWeSawTheProg.size > 200) return;
     const stringified = stringify(program);
     const latestSeen = latestPhaseWeSawTheProg.get(stringified);
@@ -395,7 +400,7 @@ export function compileVariantNoPacking(
 
   while (!queue.isEmpty()) {
     const state = queue.dequeue();
-    const phase = language.phases[state.startPhase];
+    const phase = phases[state.startPhase];
     const warnings = [...state.warnings];
 
     function addWarning(x: Error, isGlobal: boolean) {
--- a/src/common/compile.ts
+++ b/src/common/compile.ts
@@ -280,7 +280,11 @@ export function compileVariantNoPacking(
   language: Language,
 ): CompilationResult {
   const phases = language.phases;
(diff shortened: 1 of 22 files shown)


jared-hughes:
https://github.com/polygolf-lang/polygolf/tree/interpreter is an old branch that did some interpreter work. I suspect that eval'ing some JS code (your PR) is better for performance, but hand-rolling the interpreter may be better to ensure that more behavior is consciously/explicitly decided upon. I wonder if compiling to WebAssembly would give the best of both worlds here (at the cost of taking more time to write).

jared-hughes:
Just stumbled across https://www.assemblyscript.org/. Looks cool, not sure if relevant.

MichalMarsalek:
Not sure how much more work there was left to do (+ how much work would it be to merge - 157 commits behind is more than half of main's commits).
I imagine the tree walking interpreter could be more than 100 -1000 times slower than the output in say C, which would make it infeasible if user wants to compile a solution that uses an algo that takes seconds to run on code.golf.
One think I thought was cool about your work was the idea of checking (and perhaps narrowing) the types of nodes when interpreting. I thought this was something that we need a custom interpreter for, but now I realize it could be done by more (albeit much more) instrumentations on whatever language we decided to transpile to and run. Not sure how much would basically wrapping every value in a call to some instrumentation function slow down the execution.

src/common/compile.ts
```diff
@@ -242,6 +243,10 @@ export function compileVariant(
   options: CompilationOptions,
   language: Language,
 ): CompilationResult {
+  if (options.level !== "nogolf")
+    try {
+      getOutput(program); // precompute output
+    } catch {}
```
jared-hughes, line 249: What's the point here? I'd think the `hardcode` plugin would end up running `getOutput` anyways, which caches the result.

src/interpreter/index.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+  function write(x: string) {
+    output += x;
+  }
+  const start = Date.now();
+  function instrument() {
+    if (Date.now() - start > 500)
+      throw new PolygolfError("Program took too long to interpret.");
+  }
+  eval(jsCode.result);
```
jared-hughes, line 81: There's no guarantee that esbuild will preserve the names of `print` etc. I've seen it change `func` to `func1` if there's another one in global scope.

An option to closure in the variables is:
```
(new Function("print", "write", "instrument", jsCode.result))(print, write, instrument);
```

src/interpreter/index.ts
```diff
@@ -0,0 +1,84 @@
...
+import { addVarDeclarations } from "../plugins/block";
+
+const javascriptForInterpreting = {
+  ...javascriptLanguage,
+  phases: [
+    ...javascriptLanguage.phases,
+    required(addVarDeclarations, {
+      name: "instrumentInNogolf",
+      visit(node, spine, context) {
+        if (context.options.level === "nogolf") {
```
jared-hughes, line 23: Why does it need to check for `"nogolf"`?

src/languages/python/python.test.md
```diff
@@ -161,14 +161,14 @@ a-=a*a
 ```polygolf
 $a:0..oo <- 1;
 if ($a != 0) {
-    println_int $a;
+    $a <- 2;
 };
 $a <- 1;
 ```
 
 ```python
```
jared-hughes, line 169: Maybe we could support `python nohardcode` to specifically exclude the hardcode plugin.

src/common/compile.ts
MichalMarsalek: > What's the point here? I'd think the `hardcode` plugin would end up running `getOutput` anyways, which caches the result.

In the past we asserted that there's always only a one compilation going on at a time, but this would run a compilation inside a compilation.

src/interpreter/index.ts
MichalMarsalek: It doesn't, that's a leftover from when I was modifying javascript directly. Good catch.

src/languages/python/python.test.md
MichalMarsalek: > Maybe we could support `python nohardcode` to specifically exclude the hardcode plugin.

Yes, I was thinking `no:pluginName`. `no:inlineVariables` would be useful too.

PR #265, opened 2023-11-11
[WIP] Add lang: TeX

jared-hughes:
Will want to improve before merging.

Currently producing some nasty code for fibonacci:

```tex
\newcount\a\newcount\b\newcount\i\newcount\t\newcount\T\newcount~\a0 \b1 \i0
\def\c{{\ifnum\i<31 \the\a\endgraf\t\a\advance\t\b\T\t\a\b\b\T~1 \advance~\i\i~\c\fi}}\c
```

Some TODOs for correctness:
- [ ] there's excess curly braces there: Should be `\def\c{...}`, not `\def\c{{...}}`.
- [ ] Tons in the code
- [ ] Test on some solutions other than fibonacci.

Some TODOs for golfing:
- [ ] `~` should get chosen earlier by frequency
- [ ] Better 2AC conversion. Can be done at creation time, or post-hoc logic similar to `tempVarToMultipleAssignment`.
- [ ] Tons in the code

### DIFF
--- a/src/IR/IR.ts
+++ b/src/IR/IR.ts
@@ -9,22 +9,24 @@ import {
   type VarDeclarationBlock,
 } from "./assignments";
 import { type Array, type List, type Table, type Set } from "./collections";
-import {
-  type Op,
-  type Infix,
-  type ConditionalOp,
-  type FunctionCall,
-  type MethodCall,
-  type Prefix,
-  type IndexCall,
-  type KeyValue,
-  type RangeIndexCall,
-  type Function,
-  type NamedArg,
-  type ImplicitConversion,
-  type PropertyCall,
-  type Postfix,
+import type {
+  Op,
+  Infix,
+  ConditionalOp,
+  FunctionCall,
+  MethodCall,
+  Prefix,
+  IndexCall,
+  KeyValue,
+  RangeIndexCall,
+  Function,
+  NamedArg,
+  ImplicitConversion,
+  PropertyCall,
+  Postfix,
+  ScanningMacroCall,
 } from "./exprs";
+import type { FunctionDefinition } from "./functions";
 import {
   type ForRange,
   type ForEach,
@@ -41,7 +43,7 @@ import {
   type Integer,
   type Text,
 } from "./terminals";
-import { type Block, type If, type Import, type Variants } from "./toplevel";
+import type { CapturingBlock, Block, If, Import, Variants } from "./toplevel";
 import { type Type } from "./types";
 
 export * from "./assignments";
@@ -52,6 +54,7 @@ export * from "./loops";
 export * from "./terminals";
 export * from "./toplevel";
 export * from "./types";
+export * from "./functions";
 
 export interface BaseNode {
   readonly source?: SourcePointer;
@@ -90,6 +93,8 @@ export type Node =
   | ForArgv
   | If
   // Other nodes
+  | FunctionDefinition
+  | CapturingBlock
   | ImplicitConversion
   | VarDeclaration
   | VarDeclarationWithAssignment
@@ -99,6 +104,7 @@ export type Node =
   | MutatingInfix
   | IndexCall
   | RangeIndexCall
+  | ScanningMacroCall
   | MethodCall
   | PropertyCall
   | Infix
--- a/src/IR/terminals.ts
+++ b/src/IR/terminals.ts
@@ -45,6 +45,12 @@ export interface Text<Value extends string = string> extends BaseNode {
   readonly value: Value;
 }
 
+export type IDCastable = string | Identifier;
+export function castID(name: IDCastable) {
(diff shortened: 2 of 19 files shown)


src/IR/IR.ts
```diff
@@ -99,6 +104,7 @@ export type Node =
   | MutatingInfix
   | IndexCall
   | RangeIndexCall
+  | ScanningMacroCall
```
MichalMarsalek, line 107: New node kinds need to be added to parser, polygolf emitter and to parse-emit.test.ts.

src/common/emit.ts
```diff
@@ -59,10 +59,10 @@ export function containsMultiNode(exprs: readonly IR.Node[]): boolean {
 
 export class EmitError extends PolygolfError {
   constructor(expr: Node, detail?: string) {
-    if (detail === undefined && "op" in expr && expr.op !== null)
-      detail = expr.op;
+    const kind =
+      expr.kind + ("op" in expr && expr.op !== null ? `[${expr.op}]` : "");
     detail = detail === undefined ? "" : ` (${detail})`;
-    const message = `emit error - ${expr.kind}${detail} not supported.`;
+    const message = `emit error - ${kind}${detail} not supported.`;
```
MichalMarsalek, line 65: This is outdated, the only node kind with `op` is `Op` and it is never null (it is of type `OpCode`).

src/common/stringify.ts
```diff
@@ -6,7 +6,7 @@ export function stringify(x: Node, skipTargetType = false): string {
   const result = JSON.stringify(
     x,
     (key, value) =>
-      key === "source"
+      key === "source" || key === "type"
```
MichalMarsalek, line 9: Why?

src/languages/tex/plugins.ts
```diff
@@ -0,0 +1,222 @@
...
+      isExpanded: true,
+    });
+  } else {
+    throw new EmitError(node, "not integer and not string");
+  }
+}
+
+function mutatingInfixToMacros(node: MutatingInfix, spine: Spine) {
+  const { variable, right } = node;
+  const varType = getType(variable, spine.root);
```
MichalMarsalek, line 136: you pass just `spine` instead of `spine.root`

src/languages/tex/plugins.ts
```diff
@@ -0,0 +1,222 @@
...
+
+function assertIdentifier(
+  n: IR.Node,
+  detail: string,
+): asserts n is IR.Identifier {
+  if (n.kind !== "Identifier") throw new EmitError(n, detail);
+}
+
+function assertImmediate(n: IR.Node, detail: string): asserts n is Immediate {
+  if (n.kind !== "Identifier" && n.kind !== "Integer")
```
MichalMarsalek, line 220: there's `!isOfKind("Identifier", "Integer")(n)`

src/languages/tex/emit.ts
```diff
@@ -0,0 +1,125 @@
+import { type TokenTree } from "@/common/Language";
+import { EmitError, emitIntLiteral } from "../../common/emit";
+import { type IR } from "../../IR";
+import { type CompilationContext } from "@/common/compile";
+import { SPACE_ANTIGOBBLE } from "./detokenizer";
+
+// TODO-tex: somehow deal with quoting text, e.g. '#' cannot be written as-is.
+// TODO-tex: <=, >=,!= are unsupported. Plugin to convert to >,<,=
```
MichalMarsalek, line 8: golfscript does this, but there's no dedicated plugin

src/common/stringify.ts
jared-hughes: At some point I was inserting `type: intType(low,high)` annotations from the middle of a plugin that ran before `addMutatingInfix`. But then it wasn't detecting the `i <- i+1` as a mutating infix:
https://github.com/polygolf-lang/polygolf/blob/a920b88c0403bbf4df60ed9f101856dc6599a7d2/src/plugins/ops.ts#L222-L225
Pruning out `key === "type"` is no longer needed for that use case, and I suspect you want to keep in the type annotations for elsewhere.
But if this scenario comes up, the better approach might be changing `addMutatingInfix` to not use stringify to detect an identifier with the same name --- it should just scan for `node.kind === "Identifier" && node.name === lhsName`, with a separate case for index call `a[i] <- a[i] + 1`.

src/IR/terminals.ts
```diff
@@ -45,6 +45,12 @@ export interface Text<Value extends string = string> extends BaseNode {
   readonly value: Value;
 }
 
+export type IDCastable = string | Identifier;
```
MichalMarsalek, line 48: Should this be `= string | Identifier<true, string>`?

jared-hughes: No, this is for arbitrary identifiers, including names of defined functions.

PR #266, opened 2023-11-12
Flatten block after replacement causes nesting

MichalMarsalek:


jared-hughes (approved):
I've added a test that fails before this PR. Looks good

PR #267, opened 2023-11-12
Plugin improvements

MichalMarsalek:
- Allows one to use `PluginVisitor` directly instead of `Plugin` since it's just a simple wrapper.
- Adds access to a `skip` function to plugins. Once called, traversing to the replacement / to any children is skipped.
- Uses previous point to replace `if (spine.isRoot){logic}` with `context.skip("children"); logic` in several plugins.

### DIFF
--- a/src/common/Spine.ts
+++ b/src/common/Spine.ts
@@ -178,7 +178,9 @@ export class Spine<N extends IR.Node = IR.Node> {
   }
 }
 
-export type PluginVisitor<T> = <N extends IR.Node>(
+export type PluginVisitor<T = IR.Node[] | IR.Node | undefined> = <
+  N extends IR.Node,
+>(
   node: N,
   spine: Spine<N>,
   context: CompilationContext,
--- a/src/common/Spine.ts
+++ b/src/common/Spine.ts
@@ -105,7 +105,7 @@ export class Spine<N extends IR.Node = IR.Node> {
    * if the values are not used. Name inspired by Swift's `compactMap`. */
   *compactMap<T>(func: Visitor<T | undefined>): Generator<T, void, undefined> {
     let skipChildren = false;
-    function skip(kind: "replacement" | "children" | "replacementChildren") {
+    function skip(kind: "replacement" | "children") {
       skipChildren = kind === "children";
     }
     const ret = func(this.node, this, skip);
@@ -146,7 +146,7 @@ export class Spine<N extends IR.Node = IR.Node> {
   ): Spine {
     let skipReplacement = skipReplaced;
     let skipChildren = false;
-    function skip(kind: "replacement" | "children" | "replacementChildren") {
+    function skip(kind: "replacement" | "children") {
       skipReplacement ||= kind === "replacement";
       skipChildren ||= kind === "children";
     }
--- a/src/common/Spine.ts
+++ b/src/common/Spine.ts
@@ -1,5 +1,5 @@
-import { type IR, isOp, op, isOfKind, block } from "../IR";
-import type { SkipOverride, CompilationContext } from "./compile";
+import { type IR, isOp, op, isOfKind, block, type Node } from "../IR";
+import type { VisitorContext, CompilationContext } from "./compile";
 import { getChild, getChildFragments, type PathFragment } from "./fragments";
 import { replaceAtIndex } from "./arrays";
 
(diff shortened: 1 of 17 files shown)


src/common/Spine.ts
```diff
@@ -178,13 +190,19 @@ export class Spine<N extends IR.Node = IR.Node> {
...
   node: N,
   spine: Spine<N>,
   context: CompilationContext,
 ) => T;
 
-export type Visitor<T> = <N extends IR.Node>(node: N, spine: Spine<N>) => T;
+export type Visitor<T> = <N extends IR.Node>(
+  node: N,
+  spine: Spine<N>,
+  skip: SkipOverride,
```
jared-hughes, line 204: Instead of passing `skip: SkipOverride`, it would be safer (in case of a future adding more functions like `skip`) to pass in `ctx: VisitorContext` where
```ts
interface VisitorContext {
  skip: SkipOverride
}
interface CompilationContext extends VisitorContext {
  options: CompilationOptions;
  addWarning: AddWarning;
}
```
It could even be
```ts
interface VisitorContext {
  skipReplacement: () => void;
  skipChildren: () => void;
}
interface CompilationContext extends VisitorContext {
  options: CompilationOptions;
  addWarning: AddWarning;
}
```

src/common/Spine.ts
```diff
@@ -169,7 +181,7 @@ export class Spine<N extends IR.Node = IR.Node> {
         }
       }
       return curr;
-    } else if (skipReplaced) {
+    } else if (skipReplacement || skipChildren) {
       return this.replacedWith(ret);
```
jared-hughes, line 185: Looks like `skipReplaced` (function parameter) is ignored now.

src/common/Spine.ts
```diff
@@ -139,8 +144,15 @@ export class Spine<N extends IR.Node = IR.Node> {
     skipReplaced = false,
     skipThis = false,
   ): Spine {
-    const ret = skipThis ? undefined : replacer(this.node, this);
-    if (ret === undefined) {
+    let skipReplacement = skipReplaced;
+    let skipChildren = false;
+    function skip(kind: "replacement" | "children") {
+      skipReplacement ||= kind === "replacement";
+      skipChildren ||= kind === "children";
```
jared-hughes, line 151: As I understand it:
- `skipReplacement` prevents recursion into the new node (if there is a replacement)
- `skipChildren` prevents recursion into the new node (if there is a replacement) *and* into the children of the old node.
But I'm not sure about this behavior. Could we have a comment clarifying? A good place for the comment would be on the `VisitorContext` interface, if you go with that idea:
```
interface VisitorContext {
  skipReplacement: () => void;
  skipChildren: () => void;
}
```

src/markdown-tests/index.ts
```diff
@@ -72,11 +73,9 @@ export function testPlugin(
...
         applyAllToAllAndGetCounts(
           getOnlyVariant(parse(input, false)),
-          {
-            addWarning: () => {},
-            options: compilationOptionsFromKeywords(args),
-          },
-          ...plugins.map((x) => x.visit),
+          compilationOptionsFromKeywords(args),
+          () => {},
+          ...plugins.map((x) => (typeof x === "function" ? x : x.visit)),
```
jared-hughes, line 78: Is the old `Plugin` (non-`PluginVisitor`) used anymore?

src/common/Spine.ts
MichalMarsalek: Nice

src/common/Spine.ts
MichalMarsalek: It's not ignored, it's the default (initial) value of `skipReplacement`. We want to keep the old logic for skipping but additionally skip when requested from plugin. But I cannot merge the two variables, because I need to pass the original value from function parameter to the recursive calls.

src/common/Spine.ts
MichalMarsalek: Yeah sorry, it's confusing. I went trough several iterations of what it is I actually want it to do.

src/markdown-tests/index.ts
MichalMarsalek: Functions that return plugins return the proper `Plugin` type. They could return a function, but then it wouldn't be possible to include the args in the function name. But that wouldn't really matter....
However maybe we will want some more options in the Plugin interface in the future....

PR #276, opened 2023-11-23
Split & merge prints

### LINKED ISSUE #159: Split print into multiple prints
if string concatenation's only purpose is to be printed, then split each appended part into its own print statement (C, other languages with long concat)

### LINKED ISSUE #160: Merge several prints into one print

MichalMarsalek:
Closes #159.
Closes #160.


MeWhenI:
It looks like the `splitPrint` plugin isnt currently used by any langs, is that intended?

MichalMarsalek:
Ah, yes it is intended. The `splitPrint` plugin is intended for C, I don't think any of the current langs have long concatenations to warrant using this plugin.

MeWhenI:
Ah gotcha, I believe it could have value in gs since concat is 1 byte and print is 0 bytes

MichalMarsalek:
Ah, right, added it to gs.

MeWhenI (approved):

PR #279, opened 2023-11-24
fix gs print

MeWhenI:
The print mapOps replacement was clashing with other replacements in the same mapOps, fixes this. Causes cover to correctly recognize that gs has `argv_get`.

### DIFF
--- a/src/languages/golfscript/index.ts
+++ b/src/languages/golfscript/index.ts
@@ -68,13 +68,7 @@ const golfscriptLanguage: Language = {
   emitter: emitProgram,
   phases: [
     search(hardcode()),
-    required(
-      printIntToPrint,
-      implicitlyConvertPrintArg,
-      printLnToPrint,
-      printConcatToMultiPrint,
-      printToImplicitOutput,
-    ),
+    required(printIntToPrint),
     search(
       flipBinaryOps,
       equalityToInequality,
@@ -100,8 +94,11 @@ const golfscriptLanguage: Language = {
           !isSubtype(getType(node.start, spine.root.node), integerType(0)),
       ),
       replaceToSplitAndJoin,
+      implicitlyConvertPrintArg,
+      printLnToPrint,
     ),
     simplegolf(
+      printConcatToMultiPrint,
       useBuiltinAliases({ "\n": "n" }),
       alias({
         Integer: (x) => x.value.toString(),
@@ -171,6 +168,9 @@ const golfscriptLanguage: Language = {
         text_byte_reversed: (x) => infix("%", x[0], int(-1)),
         int_to_text_byte: (x) => infix("+", list(x), text("")),
       }),
+    ),
+    required(
+      printToImplicitOutput,
       addImports({ a: "a" }, (x) =>
         x.length > 0 ? assignment("a", builtin("")) : undefined,
       ),
--- a/src/languages/golfscript/index.ts
+++ b/src/languages/golfscript/index.ts
@@ -133,7 +133,6 @@ const golfscriptLanguage: Language = {
             ...(isIntLiteral()(x[0]) ? [add1(x[0]), x[1]] : [x[0], sub1(x[1])]),
           ),
       }),
-      mapOps({ print: (x) => x[0] }),
       mapToPrefixAndInfix({
         not: "!",
         bit_not: "~",
--- a/src/languages/golfscript/index.ts
+++ b/src/languages/golfscript/index.ts
@@ -29,8 +29,13 @@ import {
   removeImplicitConversions,
   printIntToPrint,
 } from "../../plugins/ops";
-import { alias, renameIdents } from "../../plugins/idents";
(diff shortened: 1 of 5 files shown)


src/languages/golfscript/index.ts
```diff
@@ -63,10 +68,15 @@ const golfscriptLanguage: Language = {
   emitter: emitProgram,
   phases: [
     search(hardcode()),
-    required(printIntToPrint),
+    required(
+      printIntToPrint,
+      implicitlyConvertPrintArg,
```
MichalMarsalek, line 73: Is there a reason these are not applied after the seearch?

src/languages/golfscript/index.ts
```diff
@@ -124,8 +133,8 @@ const golfscriptLanguage: Language = {
             ...(isIntLiteral()(x[0]) ? [add1(x[0]), x[1]] : [x[0], sub1(x[1])]),
           ),
       }),
+      mapOps({ print: (x) => x[0] }),
```
MichalMarsalek, line 136: This is not needed, right?

src/languages/golfscript/index.ts
```diff
@@ -89,8 +99,8 @@ const golfscriptLanguage: Language = {
         (node, spine) =>
           !isSubtype(getType(node.start, spine.root.node), integerType(0)),
       ),
-      implicitlyConvertPrintArg,
       replaceToSplitAndJoin,
+      useBuiltinAliases({ "\n": "n" }),
```
MichalMarsalek, line 103: This doesn't need to be applied as search, right?

src/languages/golfscript/index.ts
```diff
@@ -63,10 +68,15 @@ const golfscriptLanguage: Language = {
   emitter: emitProgram,
   phases: [
     search(hardcode()),
-    required(printIntToPrint),
+    required(
+      printIntToPrint,
+      implicitlyConvertPrintArg,
+      printLnToPrint,
+      printConcatToMultiPrint,
```
MichalMarsalek, line 75: Do you like this more than a single plugin?

src/languages/golfscript/index.ts
MeWhenI: only that they need to happen before `printToImplicitOutput`

src/languages/golfscript/index.ts
MeWhenI: oh right whoops, ill delete that

src/languages/golfscript/index.ts
MeWhenI: I believe so, since `printConcatToMultiPrint` can also golf concats that weren't generated by `printLnToPrint`, like `print (concat $a $b);`

src/languages/golfscript/index.ts
MichalMarsalek: > only that they need to happen before `printToImplicitOutput`

Yes, I meant all of the 73-76.

src/languages/golfscript/index.ts
MeWhenI: My understanding was it can be in required since it always makes the code strictly shorter.

src/languages/golfscript/index.ts
MeWhenI:
```suggestion
```

src/languages/golfscript/index.ts
MichalMarsalek: > My understanding was it can be in required since it always makes the code strictly shorter.

Yes, sorry misread it as being applied as search.

src/languages/golfscript/index.ts
MeWhenI: Then no not particularly, would there be benefit to moving them after the search?

src/languages/golfscript/index.ts
MichalMarsalek: But theoretically since it is not required to produce valid gs, it should be `simplegolf`.

src/languages/golfscript/index.ts
MichalMarsalek: Just in general if you apply the required plugins last, I think it's easier to ensure you end up with emittable program and don't have to worry about what the search plugins are doing... Also if you perform too many languages specific transforms before the main search, you might prevent some language agnostic golfing to happen.
I consider the `printIntToPrint` to be different, since that's like a preprocessing step that's gonna be included in (almost) every lang so that the rest of Polygolf doesn't even need to know that `print_int` exists.

But here, I don't think it matters much.

src/languages/golfscript/index.ts
MeWhenI: right, ill fix

src/languages/golfscript/index.ts
MeWhenI: makes sense, I'll go ahead and shift them to after the search.

PR #286, opened 2023-11-29
[WIP] More coverage

MeWhenI:
Continue to fill up op coverage
- [ ] gs
- [ ] js
- [ ] lua
- [ ] nim
- [x] py
- [ ] swift
- [ ] Tests for all

### DIFF
--- a/src/languages/python/index.ts
+++ b/src/languages/python/index.ts
@@ -15,6 +15,7 @@ import {
   type Text,
   builtin,
   isText,
+  implicitConversion,
   infix,
   list,
 } from "../../IR";
@@ -47,7 +48,11 @@ import {
   hardcode,
   listOpsToTextOps,
 } from "../../plugins/static";
-import { golfLastPrint, implicitlyConvertPrintArg } from "../../plugins/print";
+import {
+  golfLastPrint,
+  implicitlyConvertPrintArg,
+  putcToPrintChar,
+} from "../../plugins/print";
 import {
   packSource2to1,
   packSource3to1,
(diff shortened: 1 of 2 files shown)


src/languages/python/emit.ts
```diff
@@ -235,6 +235,8 @@ export default function emitProgram(
         }
         case "Prefix":
           return [e.name, emit(e.arg, prec)];
+        case "Set":
+          return ["{", joinNodes(",", e.exprs), "}"];
```
MichalMarsalek, line 239: This is a bitset which can only hold ordinal types with no more than 2^16 elements. Should probably use a hashset fro simplicity, otherwise targetType flooding ala bigint in Js might be needed.

MeWhenI: I was not aware of that, how would you specify a hashset rather than a bitset?

MichalMarsalek: > I was not aware of that, how would you specify a hashset rather than a bitset?

`["[", joinNodes(",", e.exprs), "]",".","toSet"]` plus `Set: "sets"` row in `addNimImports`. A bit hacky, but this is how tables are emitted rn.

MeWhenI: Ah, this code is for Python. I can go ahead and add that set notation to Nim though.

src/languages/golfscript/emit.ts
```diff
@@ -100,8 +100,34 @@ export default function emitProgram(program: IR.Node): TokenTree {
 
   function emitNode(expr: IR.Node): TokenTree {
     switch (expr.kind) {
-      case "Assignment":
+      case "Assignment": {
+        //a.i.@<[x]+@@)>+:a
```
MeWhenI, line 104:
```suggestion

```

PR #300, opened 2023-12-05
[WIP] Janet

MeWhenI:
82% feature coverage, 81% op coverage.

Lang features TO-DO:
- [x] add `for with step` (probably using `loop` macro)
- [x] figure out whether Janet has sets (it doesn't)
- [x] implement a few more ops (namely the right-align variants)
- [x] add mutating ops (`+=`, `-=`, `*=`, `%=`, `++`, `--`)
- [x] Appropriately implement aliasing
 
Lang-specific golfs TO-DO:
- [x] Tailor `golfLastPrint` to Janet (`println[Int]` is shorter than `print[Int]` but `print[Text]` is shorter than `println[Text]`)
- [x] Use `take` and/or `drop` over `slice` where possible
- [x] Convert `(not(nil? x))` to `(int? x)` where applicable
- [ ] Use `peg/replace-all` over `string/replace-all` where possible
- [ ] Convert if chain to `cond`
- [ ] Convert `(=(% x 2)0` to `(even? x)` and do the same for `odd?`
- [ ] Convert `(arr(-(length arr)1))` to `(last arr)`
- [ ] Use struct syntax `{}` instead of table syntax `@{}` if a Table is never mutated
- [ ] Convert `(+ (+ a b) c)` to `(+ a b c)`, same for other variadic ops


MeWhenI:
Making this non-draft since all thats left on the TO-DO is lang specific golfs, the language is fully functional.

src/languages/janet/emit.ts
```diff
@@ -0,0 +1,170 @@
...
+          36: ["36r", ""],
+        });
+      case "FunctionCall":
+        return ["(", emit(e.func), e.args.map((x) => emit(x)), ")"];
+      case "MutatingInfix": {
+        const incOrDec =
+          (e.name === "+" || e.name === "-") && isInt(1n)(e.right);
+        if (incOrDec) {
+          return ["(", e.name === "+" ? "++" : "--", emit(e.variable), ")"];
+        }
```
MichalMarsalek, line 127: This is done using a plugin for js, maybe we want to be consistent? However, the plugin creates a postfix (should it be prefix?) node, which you are not emiting.

MeWhenI: Gotcha. I can add prefix/postfix in order to work with a plugin if need be. I left prefix and infix undefined just for simplicity. Janet is Lisp-like and doesn't have any precedence rules, so it seemed the simplest just to use FunctionCall for everything since its naturally variadic.

MeWhenI: At some point I'll probably go back and do the same with golfscript since theres no real reason for a distinction between prefix and infix in a polish notation language.

src/languages/janet/emit.ts
```diff
@@ -0,0 +1,170 @@
+import {
+  EmitError,
+  emitIntLiteral,
+  emitTextFactory,
+  joinTrees,
```
MichalMarsalek, line 5: Just use the array of trees directly instead of interleaving with `""` tokens?

src/languages/janet/plugins.ts
```diff
@@ -0,0 +1,8 @@
+import { type Spine } from "../../common/Spine";
+import { implicitConversion, isOp, type Node } from "../../IR";
+
+export function implicitlyConvertConcatArg(node: Node, spine: Spine) {
+  if (isOp("int_to_dec")(node) && isOp("concat[Text]")(spine.parent!.node)) {
```
MichalMarsalek, line 5: `node` can theoretically be root so `spine.parent` might be undefined

src/languages/janet/emit.ts
MichalMarsalek: Yes, adding I agree that adding the prefix/infix/postfix nodes to Janet doesn't make much sense. But I think the logic of detecting that something is in fact inc/dec op should be centralised.

src/languages/janet/index.ts
```diff
@@ -0,0 +1,199 @@
...
+        "concat[List]": (x) => func("array/concat", list([]), ...x),
+        "find[byte]": (x) => func("string/find", x[1], x[0]),
+        "ord[byte]": (x) => op("ord_at[byte]", x[0], int(0n)),
+        "slice[byte]": (x) =>
+          rangeIndexCall(x[0], x[1], op("add", x[1], x[2]), int(1n)),
+        "slice[List]": (x) =>
+          rangeIndexCall(x[0], x[1], op("add", x[1], x[2]), int(1n)),
+      }),
+      addMutatingInfix({ add: "+", sub: "-", mul: "*", rem: "%" }),
+      mapTo(func)({
```
MichalMarsalek, line 120: `mapToPrefixAndInfix` used in other langs has special casing for unary and binary `-` so that `((-1) * $a)` is emitted as `-a` and `((-1 * $a) + $b)` is emitted as `b-a` which won't happen using the `mapTo(func)({...})` mapping. (also, the op name mapping includes `neg` but is missing `sub`.)

src/plugins/ops.ts
```diff
@@ -236,14 +237,22 @@ export function addMutatingInfix(
...
 
-export function addPostfixIncAndDec(node: Node) {
-  if (
-    node.kind === "MutatingInfix" &&
-    ["+", "-"].includes(node.name) &&
-    isInt(1n)(node.right)
-  ) {
-    return postfix(node.name.repeat(2), node.variable);
-  }
+export function addPostfixIncAndDec(
```
MichalMarsalek, line 240: Ig this should just be named `addIncAndDec`.

src/languages/janet/index.ts
MichalMarsalek: In Janet, `(- a b c d)` evaluates to `a-b-c-d` which we should be able to use.

MeWhenI: I got Janet working with `mapToPrefixAndInfix` (renamed to `mapUnaryAndBinary`), I'm having trouble combining that with getting Janet to do variadic emits like `(- a b c d)` but I'll add that to the golf goals for the lang

MichalMarsalek:
Nice generalisation of the `mapToPrefixAndInfix` plugin but unless I'm missing something, you don't want the chaining? Doesn't this compile `(+ a b c)` to `(+(+ a b)c)`? (cannot try it rn)

MeWhenI:
Yes, using toBinaryChain gets the functionality of replacing `(* -1 x)` to `(- x)` but kills variadic ops. I could further generalize it with a `preserveVariadicOps` boolean flag, or if thats stretching too far from the original design of the plugin I can just create a separate one specific to Janet/Lisp-likes

MichalMarsalek:
I would just ditch the chaining altogether and change `Infix` node to accept variadic number of children. Langs can easily emit with `joinTrees`.

PR #313, opened 2023-12-06
Fix negative int literal as bigint in js

### LINKED ISSUE #303: Javascript improper int type mixing
[Playground link](https://polygolf-lang.github.io/?source=for%20%24a%201%20100%20%7B%0A%20println%20((pow%202%20%24a)%20%2B%201)%3B%0A%20println%20((pow%202%20%24a)%20-%201)%3B%0A%7D%3B&language=Javascript&objective=bytes&getAllVariants=false&compile=true)

**Expected output**

```js
B=BigInt
for(a=1;a<100;a++)print(1n+2n**B(a)),print(2n**B(a)-1n)
```

**Comment**
The first print correctly adds a BigInt literal where needed, but the second print does not.
JS throws `TypeError: Cannot mix BigInt and other types, use explicit conversions`

MichalMarsalek:
Fixes #303

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -7,15 +7,10 @@ import compile, { debugEmit } from "./common/compile";
 import { PolygolfError } from "./common/errors";
 import languages, { findLang } from "./languages/languages";
 import { EmitError } from "./common/emit";
-import { id, int, integerType, op } from "./IR";
-import { getType } from "./common/getType";
 
 const languageChoices = [
   ...new Set(languages.flatMap((x) => [x.name.toLowerCase(), x.extension])),
 ];
-
-const p = op("bit_xor", { ...id("a"), type: integerType(1, 100) }, int(3));
-console.log(getType(p, p));
 
 const options = yargs()
   .options({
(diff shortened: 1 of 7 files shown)


MeWhenI:
This appears to fix the case in the issue but there are still other ops where the problem persists, as all of these lines produce invalid js:

```
for $a 1 100 {
 println ((pow 2 $a) ~ 3);
 println ((pow 2 $a) | 3);
 println ((pow 2 $a) & 3);
 println ((pow 2 $a) mod 3);
};
```

MichalMarsalek:
`~`, `|`, `&` were caused by a bug in the type inference. `mod` is caused by the bigint flag being decided only based on the type of the result (which is 0..2 in this case) and not by the type of the other op's args.

src/cli.ts
```diff
@@ -7,11 +7,16 @@ import compile, { debugEmit } from "./common/compile";
...
 import { EmitError } from "./common/emit";
+import { id, int, integerType, op } from "./IR";
+import { getType } from "./common/getType";
 
 const languageChoices = [
   ...new Set(languages.flatMap((x) => [x.name.toLowerCase(), x.extension])),
 ];
 
+const p = op("bit_xor", { ...id("a"), type: integerType(1, 100) }, int(3));
+console.log(getType(p, p));
```
MeWhenI, line 18: Looks like this was left in from debugging?

PR #320, opened 2023-12-09
Slice ops

### LINKED ISSUE #294: Add `starts_with` & `ends_with` opcodes

MichalMarsalek:
- Fixes slice return type calculation
- Adds `slice_back` family of ops
- Adds `starts_with` & `ends_with` ops - Closes #294


MeWhenI:
Maybe it should be an error to attempt to compile `slice_back` where `length + index > 0`, as this will always yield an illegal index, regardless of the length of the string.

MeWhenI:
It seems like `slice` shouldn't be an alias for both `slice[Ascii]` and `slice_back[Ascii]` as it is the only op/alias that can take either a front-index or a back-index. I think for consistency there should either be another op `slice_back` that aliases `slice_back[Ascii]` and `slice_back[List]`, or every `slice` opcode should be an alias for its associated `slice_back` opcode.

MeWhenI:
Compiling `print (at_back[List] (list 1 2 3) -1);`.
 Lua uses an invalid index:
```lua
print({1,2,3}[0])
```
Javascript produces:
```js
print`undefined`
```

MeWhenI:
I couldn't find an easier way to suggest a change on a file that wasn't edited in this PR so I made PR #323 to fix a small bug with `set_at` in golfscript.

MichalMarsalek:
> It seems like `slice` shouldn't be an alias for both `slice[Ascii]` and `slice_back[Ascii]` as it is the only op/alias that can take either a front-index or a back-index. I think for consistency there should either be another op `slice_back` that aliases `slice_back[Ascii]` and `slice_back[List]`, or every `slice` opcode should be an alias for its associated `slice_back` opcode.

I don't understand why it shouldn't be. It's not the only alias that can take forwards or backwards index, `@` can too.

MeWhenI:
> I don't understand why it shouldn't be. It's not the only alias that can take forwards or backwards index, `@` can too.
To me it seems intuitive that if `slice` can take a negative index, so could any typed version of `slice` like `slice[Ascii]` or `slice[byte]`.

MichalMarsalek:
> Compiling `print (at_back[List] (list 1 2 3) -1);`. Lua uses an invalid index:
> 
> ```lua
> print({1,2,3}[0])
> ```
> 
> Javascript produces:
> 
> ```js
> print`undefined`
> ```

This is not related to this PR, the output is the same in main. It seems to be fixed in #321 (correct output for Lua & at least an error for Javascript).

MeWhenI:
> This is not related to this PR, the output is the same in main. It seems to be fixed in #321 (correct output for Lua & at least an error for Javascript).

Ok gotcha, I had commented here because I found it while testing the slice_back op and used at_back as a minimum reproducible example.

MeWhenI:
Besides that it looks good

PR #321, opened 2023-12-16
Pure ops

MichalMarsalek:
Removes all ops that mutate variables & adds pure versions. The new idiomatic way to replace an item in a list is `$list @<- $index $value;` (which is possible thanks to slightly loosening the grammar).
Removes `MutatingInfix` node and the `addMutatingInfix` plugin is generalised so that assigning a result of an op to the same variable as the first argument of that op (or any argument if commutative) can be treated as mutation in the target lang regardless of the syntax.
The op mapping to target langs is refactored to be more uniform, more declarative, more explicit but hopefully still general enough. Also, it's now easier to emit variadic ops as either variadic or binary depending on the op and the lang, regardless of syntax. Fixes variadic ops emit in Janet and couple of bugs along the way.


src/languages/javascript/index.ts
```diff
@@ -199,62 +209,82 @@ const javascriptLanguage: Language = {
         int_to_bool: (x) => implicitConversion("int_to_bool", x[0]),
         "contains[Table]": (x) => infix("in", x[1], x[0]),
       }),
-      mapTo((name: string, [obj, ...args]) => method(obj, name, ...args))({
+      mapMutationTo.prefix({
```
MeWhenI, line 212: js doesn't seem to be generating prefix mutations right now, compiling
```
$x:0..10 <- 1;
$x <- (+ $x 1):0..10;
```
produces
```js
x=1
x+=1
```

MichalMarsalek: Fixed.

MeWhenI:
Compiling
```polygolf
$t <- (list 1 2 3);
set_at[List] $t 2 1;
``` to Janet produces a deprecation warning to use `with_at[List]` rather than `set_at[List]`.  Compiling the same code using `with_at[List]` produces `emit error - Op (with_at[List]) not supported.`

**Comment by MeWhenI** [comment]
I see operator mappings for `with_at` in all languages, but
```
$t <- (list 1 2 3);
$t @<- 2 1;
```
throws emit errors in every language except Janet.

MichalMarsalek:
> Compiling
> 
> ```
> $t <- (list 1 2 3);
> set_at[List] $t 2 1;
> ``` to Janet produces a deprecation warning to use `with_at[List]` rather than `set_at[List]`.  Compiling the same code using `with_at[List]` produces `emit error - Op (with_at[List]) not supported.`
> ```

This is expected behaviour, but perhaps the message should be clearer and explain that the `with_at[List]` op should be used as a RHS of an assignment and not as a lone statement?

MichalMarsalek:
> I see operator mappings for `with_at` in all languages, but
> 
> ```
> $t <- (list 1 2 3);
> $t @<- 2 1;
> ```
> 
> throws emit errors in every language except Janet.

This is fixed.

PR #323, opened 2023-12-20
Slice op bugfix

MeWhenI:
Fix index assignment being broken at index -1 in gs in #320


MichalMarsalek:
So it's only a problem when the index is the -1 literal, but not when it's another node that evaluates to -1?

MeWhenI:
> So it's only a problem when the index is the -1 literal, but not when it's another node that evaluates to -1?

ah no thats right, it'll also need to work for any node that evaluates to -1, so a special case isn't sufficient.

PR #336, opened 2023-12-30
Fix #306

### LINKED ISSUE #306: Lua bad conditional typing
[Playground link](https://polygolf-lang.github.io/?source=%24a%20%3C-%20(%40%200)%3B%0Aprint%20(..%20%22Hello%20%22%20(conditional%20(1%20%3E%201)%20%22%20World%22%20%22%22))%3B&language=Lua&objective=bytes&getAllVariants=false&compile=true)

**Expected output**

```
a=arg[1]
print("Hello "..(1>1 and" World"or""))
```

**Comment**
`/usr/bin/lua: stdin:2: attempt to concatenate a boolean value`

MeWhenI:


MeWhenI:
I don't believe this needs to be done for any ops other than concat, besides on specific per-language cases where a plugin is the better solution

Issue #328, opened 2023-12-29
Some Janet-specific golfs


MeWhenI:
This issue documents some language-specific golfs that can be added for Janet:

- Use `peg/replace-all` over `string/replace-all` where possible
- Convert if chain to `cond`
- Convert `(=(% x 2)0` to `(even? x)` and do the same for `odd?`
- Convert `(arr(-(length arr)1))` to `(last arr)`
- Use struct syntax `{}` instead of table syntax `@{}` if a Table is never mutated

PR #338, opened 2023-12-30
Refactor if-chains, add to janet

MichalMarsalek:
1 more point from #328

### DIFF
--- a/src/common/emit.ts
+++ b/src/common/emit.ts
@@ -1,4 +1,4 @@
-import { type IR, type Integer, type Node } from "IR";
+import type { If, IR, Integer, Node } from "IR";
 import { PolygolfError } from "./errors";
 import { type TokenTree } from "./Language";
 import { codepoints } from "./strings";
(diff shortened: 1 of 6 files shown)


MeWhenI:
Janet could also use `cond` for implementing `ConditionalOp`, and I imagine some other langs might also be able to pull the IfChain logic over to conditionals.

MichalMarsalek:
Ah, right. It might make sense to merge the two (If and ConditionalOp) completely, but for now, I just used the logic for ConditionalOp too.

MeWhenI:
Is there a plugin that currently exists to try inverting the condition and then swapping the alternate and consequent statements? Since if-chaining as I understand it only expands the alternate, this would be beneficial. Maybe thats for a separate issue though, PR otherwise looks good.

MichalMarsalek:
> Is there a plugin that currently exists to try inverting the condition and then swapping the alternate and consequent statements? Since if-chaining as I understand it only expands the alternate, this would be beneficial. Maybe thats for a separate issue though, PR otherwise looks good.

Such plugin should probably be added, but I wonder if there's ever a reason to nest the conditions in a "nonstandard" way in the Polygolf program?

MeWhenI:
What do you mean by "nonstandard" nesting?

MichalMarsalek:
> What do you mean by "nonstandard" nesting?

One that is not
```c
if(cond){

}
else{
  if (cond){

  }
  else{
    if(cond){
    
    }
    else{
      
    }
  }
}
```

PR #339, opened 2023-12-30
Fix #248

### LINKED ISSUE #248: Unexpected multivariant output on type errors
**Describe the bug**
There are different ways of defining a 2-variant input that should all behave the same, but they don't.

**To Reproduce**
Consider these three Polygolf programs:
1.
```
{
    $x <- 0;
    print_int $x;
/
    $x <- "x";
    print_int $x;
};
```
2.
```
{
    $x <- 0;
/
    $x <- "x";
};
print_int $x;
```

3.
```
$x <- {0 / "x"};
print_int $x;
```

These three programs are equivalent in theory.

**Expected behavior**
When compiled with `-l py -a`, we expect
```
x=0
print(x)

Type error. Operator 'print_int' type error. Expected [-oo..oo] but got [(Ascii 1..1)].
  5     $x <- "x";
  6     print_int $x;
        ^
```

**Actual behavior**
Code 1. output is the expected one.

Code 2. outputs
```
x=0
print(x)

x="x"
print(x)
```

Code 3. outputs
```
x=0
print(x)

Type error. Cannot assign (Ascii 1..1) to 0..0.

    ^
```

MichalMarsalek:
Fixes #248 
This works, but maybe there are better solutions like clearing all the caches between processing of different variants.

### DIFF
--- a/src/common/expandVariants.ts
+++ b/src/common/expandVariants.ts
@@ -15,7 +15,7 @@ export function expandVariants(program: IR.Node): IR.Node[] {
   const n = numVariants(program);
   if (n > 16)
     throw new Error(`Variant count ${n} exceeds arbitrary limit. Giving up`);
-  return allVariantOptions(program);
+  return allVariantOptions(program).map((x) => structuredClone(x));
 }
 
 export function getOnlyVariant(program: IR.Node): IR.Node {


MeWhenI:
Would you want Jared to look at this before merging?

PR #343, opened 2024-01-01
Ranges and for loops

### LINKED ISSUE #33: Make syntax for range types and for looping over ranges consistent
- Remove the current syntax for loops over ranges: `for $i 0 100`
- Introduce `for $i 0..99` for inclusive upper bound and `for $i 0..<100` for exclusive upper bound instead
- Consider changing the `..` alias of `text_concat` to something else (`++` or `<>`...)
- Make sure `0..99` is parsed correctly in both type annotation context and loop context.
- Make sure the bounds can be any expression, including variants (`for $i {0 / 1}..<201 {...`)

### LINKED ISSUE #127: Range op
Add (backend only) `range` op and a plugin converting `ForRange` to `ForEach` over the range op.
Besides less repetition in the Python and Swift emitters it means that the `range` can be aliased.

MichalMarsalek:
Big refactor of integer ranges and loops. The main motivation is to simplify the AST, reduce repetition and special casing in plugins and emitters and to simplify the control flow analysis in the future.

- Closes #33 
- Closes #127 
- Renames concat alias to `+` (`..` is understood, but emits deprecation warning (this was kind of tricky to implement since it needs type info))
- Removes `ForRange`, `ForDifferenceRange` nodes
- Adds `range_incl` (alias `..`), `range_excl` (alias `..<`) as frontend and  `range_diff_excl` as backend opcodes to fill in the functionality
- Adds `text_to_list[Ascii]`, (alias `text_to_list`) `text_to_list[byte]`, `text_to_list[codepoint]` opcodes, to be used for iterating over characters.
- Removes `ForEachKey` & `ForEachPair`. They were not used and I'm not sure target langs would agree on iteration order. If we decide to add this functionality later, we can do so via a `table_keys_list` opcode over which we would iterate using a `ForEach` = `for`.
- Adds default values for opcodes. These are used when parsing Polygolf and encountering less arguments than the nominal arity and also by `mapOps` plugin, if the target lang shares the defaults with Polygolf.
- Adds `Cast` node. This node is used when explicit conversion from/to a type that doesn't exist in Polygolf is needed in the target language.

Polygolf's `for` now means a for each loop over list items and it should be used for iterating over a range as well, similar to Python.
`for $i $a $b {};` should now be written `for $i ($a ..< $b) {}`. Old form is understood (as implicit `range_excl`), but emits a deprecation warning. `for $i 10 {};` and `for 10 {};` forms are preserved without a warning.
`for $i $text {};` is a syntax sugar for `for $i (text_to_list $text) {};`, `for[byte] $i $text {};` is a syntax sugar for `for $i (text_to_list[byte] $text) {};` similar for`[codepoint]`.


Steffan153:
I think several of the language-specific tests need updated, as they still have the old `for $i $a $b` in them

PR #357, opened 2024-02-20
Add Clojure

Steffan153:
Many golfing optimizations can be made, but it works, including all of my https://code.golf/golfers/steffan-polygolf solutions.


src/languages/clojure/index.ts
```diff
@@ -0,0 +1,196 @@
...
+            a,
+          ),
+      }),
+    ),
+    simplegolf(implicitlyConvertConcatArg),
+    required(
+      mapOps({
+        argv: () => builtin("*command-line-args*"),
+
+        // TODO: vec not necessary if already a vec
```
MichalMarsalek, line 93: Perhaps this should turn into an issue? I will try to do the same for TODOs that are already in main.

src/languages/clojure/index.ts
```diff
@@ -0,0 +1,196 @@
...
+        int_to_Hex_aligned: (a, b) =>
+          func(
+            "format",
+            op["concat[Text]"](text("%0"), intToDecOpOrText(b), text("X")),
+            a,
+          ),
+      }),
+    ),
+    simplegolf(implicitlyConvertConcatArg),
+    required(
```
MichalMarsalek, line 89: There are some opcodes that could be mapped using `mapOpsTo.func` instead of the general `mapOps`.

src/languages/clojure/index.ts
```diff
@@ -0,0 +1,196 @@
...
+        append: (a, b) => func("conj", a, b),
+
+        "at[argv]": (a) => op["at[List]"](builtin("*command-line-args*"), a),
+        "at[codepoint]": (a, b) => func("str", func("nth", a, b)),
+        "contains[Text]": (a, b) => func("clojure.string/includes?", a, b),
+        "contains[Table]": (a, b) => func("contains?", a, b),
+        "ord[codepoint]": (a) => op["ord_at[codepoint]"](a, int(0n)),
+        pow: (a, b) => func("int", func("Math/pow", a, b)),
+      }),
+      mapOps({
```
MichalMarsalek, line 103: I think this `mapOps` call could be combined with the previous one?

src/languages/clojure/index.ts
```diff
@@ -0,0 +1,196 @@
...
+        "at_back[List]": (a, b) =>
+          isInt(-1n)(b) ? func("last", a) : undefined,
+      }),
+      mapBackwardsIndexToForwards({
+        "at_back[Ascii]": "size[Ascii]",
+        "at_back[codepoint]": "size[codepoint]",
+        "at_back[List]": "size[List]",
```
MichalMarsalek, line 110: `at_back[List]` is already mapped using the previous plugin

Steffan153: But it returns `undefined` unless it's -1, so this has to be implemented too.

MichalMarsalek: > But it returns `undefined` unless it's -1, so this has to be implemented too.

yes, mb

src/languages/clojure/index.ts
Steffan153: Sure, this should be an issue. I'm not sure how to go about it, but we would need a more complicated system of keeping track of the types of everything (lang-specific types, not Polygolf types).

MichalMarsalek: Maybe something similar we use in javascript for int vs bigint?

PR #368, opened 2024-03-07
Add Coconut

MichalMarsalek:
Improvements over Python:
- `os.sys` instead of `import sys`
- unicode operators for chars objective


Steffan153:
One of the main other things to do is `f(x,y,...)` is `f x y ...` if all `x,y,...` are numbers, variables, or exponentiated variables (`f x**2`).

https://coconut.readthedocs.io/en/latest/DOCS.html#implicit-function-application-and-coefficients

Steffan153 (approved):

PR #373, opened 2024-03-11
Cloning

### LINKED ISSUE #347: Standardize array copy behavior
[Playground link](https://polygolf-lang.github.io/?source=%24a%20%3C-%20(list%201%202%203)%3B%0A%24b%20%3C-%20%24a%3B%0A%24c%20%3C-%20(%40%200)%3B%0A%24b%20%40%3C-%202%202%3B%0Aprint%20(%40%20%24a%202)%3B%0Aprint%20(%40%20%24b%202)%3B&language=all&objective=bytes&getAllVariants=false&compile=true)

**Comment**
There is variance between langs of whether array assignment makes a deep copy or a shallow copy. Most langs currently make shallow copies. Which behavior should be default, and should the alternate be supported in some way?

MichalMarsalek:
Closes #347 by cloning anything that doesn't have a value semantics in the target lang.

### DIFF
--- a/src/languages/janet/index.ts
+++ b/src/languages/janet/index.ts
@@ -52,7 +52,7 @@ import {
   truncatingOpsPlugins,
 } from "../../plugins/arithmetic";
 import { forArgvToForEach, forEachToForRange } from "../../plugins/loops";
-import { alias, renameIdents } from "../../plugins/idents";
+import { alias, clone, renameIdents } from "../../plugins/idents";
 import { assertInt64 } from "../../plugins/types";
 import { implicitlyConvertConcatArg } from "./plugins";
 import { applyIf } from "../../plugins/helpers";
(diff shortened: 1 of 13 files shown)


src/languages/janet/index.ts
```diff
@@ -212,6 +212,17 @@ const janetLanguage: Language = {
...
       }),
+      clone((node, type) => {
+        if (["boolean", "integer", "text"].includes(type.kind)) {
+          return node;
+        }
+        if (
+          type.kind === "List" &&
+          ["boolean", "integer", "text"].includes(type.member.kind)
+        ) {
+          return func("array/slice", node, int(0), int(-1));
```
Steffan153, line 223: `@[;a]` is much shorter.

Steffan153: Deep copy: `(thaw(freeze a))`

MichalMarsalek: Thanks, I was searching for 10 minutes and wasn't able to find anything.

Steffan153 (approved):

PR #377, opened 2024-03-13
Add coco function call syntaxes

### LINKED ISSUE #370: Coconut golfs
- different forms of function applications
  Could be implemented similar to how `f(x)` => `f x` & `f(x)` => `x.f` is done for Nim.
  https://github.com/polygolf-lang/polygolf/blob/c849d1f8b3e2cfc39f443c478cf3bd3f3e434df7/src/languages/nim/plugins.ts#L89
  - [Implicit application/multiplication](https://coconut.readthedocs.io/en/latest/DOCS.html#implicit-function-application-and-coefficients)
  - [Pipes ](https://coconut.readthedocs.io/en/latest/DOCS.html#pipes)
  - [Infix functions](https://coconut.readthedocs.io/en/latest/DOCS.html#infix-functions)

MichalMarsalek:
Closes #370 
- [x] Implicit function application
- [x] Implicit multiplication
- [x] Pipe `|>` (`↦`)
- [x] Infix function application

As a side effect I got rid of a space between a digit and a letter in Python.


Steffan153 (commented):
`0or` will not work for removing the space (interpreting as an octal literal). Might need to add that as a special case

MichalMarsalek:
> `0or` will not work for removing the space (interpreting as an octal literal). Might need to add that as a special case

Does my last commit fix all the problematic cases?

Steffan153 (approved):

PR #385, opened 2024-04-25
Generated comparisons of outputs

MichalMarsalek:
Adds another [autogenerated file](https://github.com/polygolf-lang/polygolf/blob/generated-comparisons/docs/opcodes-outputs.generated.md) to the `docs` folder.
It lists outputs for each Polygolf feature in each supported langauge.
This uses the same algorithm as `npm run cover`.


src/cover/index.ts
```diff
@@ -228,51 +284,50 @@ const opCodes: CoverTableRecipe = Object.fromEntries(
...
                         opCode === "range_excl" && i === 1
                           ? integerType(10, 10)
                           : t,
+                        forceBuiltin,
                       ),
                     ),
                   ),
-                ),
+                );
+                return lang.stmt(isLiteral(exprs[0]) ? exprs[1] : exprs[0]);
```
MichalMarsalek, line 300: On main, we were using `(1 + 1)` expressions to test the `+` opcode, but that expression is immediately reduced to `2`, which resulted in some false positives in terms of coverage. So, the failed CI check should be ignored.

